PR #10, opened 2015-01-05
Allocation testing

erbriones:
These are the set of allocation tests that have been written along with more tests being added.


erbriones:
Finished adding initial set of tests.

PR #11, opened 2015-01-29
Jchansen/break out serializers

jchansen:
Break out serializers into their own files.  Remove api/serializers.py


jchansen:
Notes from review with Evan:

1. (action) Fields should either be broken out into their own folder or, if there's only one serializer that uses them, consumed into that serializer's module.

2. (action) Regarding the files that are not serializers or fields (get_context_user, get_projects_for_obj), need to a) review what they actually do, and b) consider moving them into __init__.py.

3. (see #4 for conversation) Need to review how imports should be specified in files.  They are wrong right now.  Need to make them right (3rd party, 3rd party us, Django modules, app modules, etc.)

4. (discussion to set coding conventions) Regarding the relative imports, we need to discuss as a team what our coding conventions are going to be.  If we decide relative imports are a no-no, then they need to be changed to absolute, otherwise we need to decide on scope of relative imports being okay.

jmatt:
Nice.

Coding style only works if Steve complies. Evan and I stay relatively PEP8 Compliant.

One use case for relative imports is to simplify (shorten) imports when the structure is deeply nested. We're not especially deeply nested right now. Another reason is ambiguity with names. Such as import client. Which could refer to a number of modules. Absolute and relative imports are useful in those situations. Have we seen ambiguity in naming and modules with the API? I could see that being a reason to use relative imports. In the past I use them exceptionally to resolve ambiguity. But maybe there is a case for using them?

See the PEP: https://www.python.org/dev/peps/pep-0328/


JMatt

arleigh16:
How do I get off of this mailing list?
 Arleigh Birchler
"Wild Azalea Home"

** Sent from my Intergalactic Star Cruiser **

arleigh16:
Since I cannot find a way to get off of this mailing list I am just marking the messages I receive as spam.
 Arleigh Birchler
"Wild Azalea Home"

** Sent from my Intergalactic Star Cruiser **

steve-gregory:
Hello Arleigh, you can change your permission here:

https://github.com/iPlantCollaborativeOpenSource/atmosphere/watchers

and select 'Unwatch' in the Top Right corner.

steve-gregory:
As for @jmatt comment I will do more to ensure code pushed in previous branches is PEP8 compliant, if we are going to develop design standards it should happen out of band of a github branch

Cheers all

jchansen:
Agree with @steve-gregory that the discussion is best had out of band.  Also, let's keep things positive.  It's important to remember that we succeed or fail as a team, not as individuals.

PR #12, opened 2015-01-31
Jchansen/upgrade to drf3

jchansen:
Upgrade v1 API to Django Rest Framework 3

### DIFF
--- a/api/serializers/cleaned_identity_serializer.py
+++ b/api/serializers/cleaned_identity_serializer.py
@@ -4,12 +4,12 @@ from rest_framework import serializers
 
 class CleanedIdentitySerializer(serializers.ModelSerializer):
     created_by = serializers.CharField(source='creator_name')
-    credentials = serializers.Field(source='get_credentials')
-    id = serializers.Field(source='uuid')
-    provider_id = serializers.Field(source='provider_uuid')
-    quota = serializers.Field(source='get_quota_dict')
-    allocation = serializers.Field(source='get_allocation_dict')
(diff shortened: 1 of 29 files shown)


jchansen:
I'll review this with @steve-gregory on Monday.  I want to walk through each of the files and talk about what changed with DRF3, as well sanity check that there aren't side-effects that I couldn't see for any of the changes.  I test with both Airport and Tropo on D-D release, launched volumes, instances, terminated, suspended, add/delete projects, everything I could think of.  I worked out any kinks I could find.

api/serializers/cleaned_identity_serializer.py
```diff
@@ -4,13 +4,13 @@
 
 class CleanedIdentitySerializer(serializers.ModelSerializer):
     created_by = serializers.CharField(source='creator_name')
```
jchansen: Steve says: "This should be a ReadOnlyField"

jchansen:
Reviewed with Steve.  He says LGTM, so going to merge it in.  A couple changes since the review:

1. Updated the created_by field in CleanedIdentitySerializer to be read only.

2. Fixed the internal DRF2 references in ProjectsField to be compatible with the DRF3 API.  This should be refactored not to use internal API at all (Evan suggests over-riding the constructor as a better alternative), but I'm going to leave it alone for now since it's only used in Tropo and the v2 API will replace it soon.  Plus DRF3 may have a more natural approach for accomplishing the goal.  If so, we'll use that, if not, we'll rewrite ProjectsField to not be naughty.

PR #13, opened 2015-02-06
Dd/fix switch selected identity bug

jchansen:
1. Reverted commit 2f296c97f32ab98fca96f48c1606281602e63f1b due to massively breaking the API

2. Removed the MachineRequestStaffList and MachineRequestStaff imports in urls.py as they don’t exist and were throwing errors on boot

3. fixed the bug that was preventing users from being able to change selected identity in v1/profile

### DIFF
--- a/api/profile.py
+++ b/api/profile.py
@@ -72,7 +72,7 @@ class Profile(APIView):
         """
         user = request.user
         profile = user.userprofile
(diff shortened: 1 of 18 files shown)


jchansen:
@steve-gregory Can you verify this works on your machine as well before we merge it in?

jchansen:
Steve says "LG2m!"  Merging!

PR #14, opened 2015-02-12
Jchansen/upgrade/api v2

jchansen:
can reuse database between test runs (only need to rebuild after a
migration)


jchansen:
After reviewing the initial pull request with @steve-gregory, the following modifications were agreed to and acted upon:

1. Serializers were combined into groups, which initially resulted in circular dependencies, but this was resolved by factoring the grouping under detail and summary serializers.  All 50 v2 API tests are passing again.

2. Removed all references to Django Debug Toolbar.  This could/should go into a different settings file that's loaded on request (if that behavior is desired) and inherits from the actual atmosphere.settings file.  The reasoning for removing the Debug Toolbar is that while it only runs in DEBUG mode, it has a severe negative impact on performance, typically resulting in API calls taking 2-3 times longer than normal.  We'll find a better way to include it that's less obtrusive.

3. I removed the InstanceAction and VolumeAction models, serializers and views.  While these models can/will be beneficial to Atmosphere, they shouldn't be included before we discuss them, and ultimately belong in their own pull request.

4. Removed the ApplicationTag and InstanceTag models.  Both of these models simply exposed already existing tables in the database, so that they could then be exposed through the API, but they also need to be discussed before including.

5. Moved the django-nose, django-filter, and djangorestframework-filters packages into the appropriate location in requirements.txt.

jchansen:
@steve-gregory Take a look at the changes and let me know if the current setup meets your expectations.  If yes I'll merge in (or you can push the button yourself).

PR #15, opened 2015-03-02
Jchansen/v2 fixes

jchansen:
Not intended to be merged yet, just prepping.  Need to test on Alom first.

1. /images?search=[words] searches across image name/description and tag name/description (mirrors v1 behavior)
2. Fixed v2 serializers to work with instance_sources
3. Made various modifications to v2 serializers to support Tropo use cases during migration to v2 (added filters, fields, etc.)
4. Added ProjectInstance endpoint and ProjectInstance model that **exposes** the already existing project_instances table - I simply needed to be able to query it.
5. Added ProjectVolumes endpoint and ProjectVolume model that **exposes** the already existing project_volumes table - I simply needed to be able to query it.


jchansen:
Notes from PR Review with @steve-gregory 

1. The serializers are returning to much information, and we need to strip them down farther.  This is partly due to some comprises being made to facilitate the v2 migration on the UI side, but is not prudent long term.  We should (IMO) make this a gradual effort in EE to continue stripping down the amount of information returned and exposing query endpoints as needed to facilitate that goal.  The most obvious place this manifests itself is the fact that RelatedFields are being specified inside various detail serializers instead of in the /fields module.  This is because putting them in /fields us causing circular dependencies, but root cause/smell is too many levels of depth on a few of the serializers.  This will be trimmed down gradually over EE until all the related fields can be moved into /fields and re-used across serializers.

2. The "queryset=" field in the DRF's RelatedSerializers is primarily used to determine what to show in the options dropdown of the DRF interface and the primarily performance bottleneck during rendering of the interface.  The problem is the field is static in DRF - it can't modified at run-time without overriding some properties/methods for every serializer that needs it (especially the 'choices' property), so there's no easy way to limit the subset of information that's displayed to the context of the user.  This means that in many views, the options dropdown has hundreds or thousands of options.  To get around this in the short-term and make the DRF interface performant, the queryset in all related fields will be set to <Model>.objects.none(), and the get_queryset method (the one actually used by serializer for core use) will be set to <Model>.objects.<whatever queryset filter you actually need>.  This will cause the DRF views to render quickly while still behaving as designed.

3. Related to 2, we need to look at a way to share querysets between RelatedFields in serializers and ViewSets to prevent code duplication.

4. Some of the Serializers (like details/instance.py) are using a Tuple syntax for multiple imports from the same module instead of the \ syntax.  @steve-gregory and I both like this and are interested in adopting it when imports exceed 80 characters instead of using backslashes.  We should discuss as a team if that's a convention we want to adopt as we continue refining and defining out coding conventions.

5. Steve asked me to add "through='<Model>'" fields to Project, Instance and Application models now that we've formally exposed the auto-generated "through" fields as models (ProjectInstance, ProjectVolume, InstanceTag and ApplicationTag).  Because the models are now explicit, if we ever changed the table_name the relationships like project.instances would break without that "through" field specified.  That change was implemented in the last commit of this PR.

PR #17, opened 2015-04-22
Add maintenance middleware

jchansen:
[lets demo this to team before merging]

This adds middleware to Atmosphere that blocks all API requests from non-staff when Atmosphere is under maintenance.

1. If user is staff, does nothing
2. If user is not staff, gets active maintenance records that have no assigned provider where disable_login is true.
3. If a record matches that description, return a 503 SERVICE UNAVAILABLE status code along with a JSON object that contains the message from the first maintenance record.

### DIFF
--- a/atmosphere/maintenance_middleware.py
+++ b/atmosphere/maintenance_middleware.py
@@ -1,0 +1,15 @@
+from django.http import JsonResponse
+
+
+class MaintenanceMiddleware(object):
+    """
+    This middleware returns a 503 when the app is under maintenance for non-admin users
+    """
+    
+    def process_request(self, request):
+        content = {
+            'message': 'Atmosphere is currently under maintenance'
+        }
+        response = JsonResponse(content)
+        response.status_code = 503
+        return response
--- a/atmosphere/maintenance_middleware.py
+++ b/atmosphere/maintenance_middleware.py
@@ -7,9 +7,12 @@ class MaintenanceMiddleware(object):
     """
     
     def process_request(self, request):
-        content = {
-            'message': 'Atmosphere is currently under maintenance'
-        }
-        response = JsonResponse(content)
-        response.status_code = 503
-        return response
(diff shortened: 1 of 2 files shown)


erbriones:
Generally looks good but few issues that will need to be addressed. 
- Allowing a staff users to login if a maintenance record is active (django admin).
- Enabling maintenance for only a specific provider

jchansen:
**Staff users**: This has no affect on staff users.  The first line checks if the user is staff, and if they are it exits without doing anything.

**Maintenance per provider**: The intent with this middleware was only to handle the case where the whole app needs to go into maintenance.  If we opt to support maintenance on a per provider basis, we'll need to better define what that means along with expected behavior, but it's outside the scope of this feature.  Right now it's considered a UI responsibility...if we opt to support it server side (return 503 when attempting to do an action on a provider in maintenance) I think that logic will need to be handled in the controllers.  Too complex for middleware to say "oh you are trying to launch, who are you launching on, let me see if they're in maintenance, yes they are, no you can't do that, 503 for you".

jchansen:
Also, it's not stated, so to be clear the use case here is that when Atmosphere goes into maintenance, users are no longer allowed to login, but logged in users can continue to launch instances because the API isn't blocking them.  This middleware is strictly intended to lock them out when we need the app to be in "admin" mode.

erbriones:
Makes sense. :hamburger: I would agree that maintenance per-cloud could be something baked into deeper. I think the expectation should be per-provider and global maintenance should be considered distinct features.

jchansen:
Discussed it as a team.  Key issues identified:

1. Middleware blocks ability to log in.  Only works when staff user is *already* logged in.

We discussed mitigating by allowing the /admin and /api-auth endpoints through, but decision was made to control permissions on a per endpoint basis instead.  Closing PR.  Will create a new branch and start implementing that approach.

PR #20, opened 2015-04-30
Add v2 quota request

erbriones:
**Add new generic request system for api v2.**

### Features
- Generic request framework
- Updating and deleted restricted to only unresolved requests
- Adds quota request endpoint
- Automated emails for quota requests
- Approved quota request will go into fail state or close state after completion

### Fixes
- Add migration for default status type in request


api/v2/serializers/details/quota_request.py
```diff
@@ -0,0 +1,161 @@
...
+    user = UserSummarySerializer(source='membership.identity.created_by', read_only=True)
+    identity = IdentityRelatedField(source='membership.identity',
+                                    queryset=Identity.objects.none())
+    provider = ProviderSummarySerializer(source='membership.identity.provider', read_only=True)
+    status = StatusTypeRelatedField(queryset=StatusType.objects.none(),
+                                    allow_null=True,
+                                    required=False)
+    quota = QuotaRelatedField(queryset=Quota.objects.all(),
+                              allow_null=True,
+                              required=True)
```
steve-gregory: I'm wondering why you subclassed PrimaryKeyRelatedField here, since its a new API model/view/serializer, you could have nested the serializers, and then override the 'create/update' fields in the serializer to handle nested saving..

steve-gregory:
Looks good to me, I give it a +1 pending that you add the required migration for StatusType

api/v2/serializers/details/quota_request.py
erbriones: Originally, the serializer didn't have a `to_internal_value` method implement so that is why it inherited from `PrimaryKeyRelatedField`. These probably should be updated to only inherit from `RelatedField`.

jchansen: I like that, changing back to 'RelatedField' for the serializers where you overrode to_internal_value.  Unless there's something else we get from 'PrimaryKeyRelatedField' that's non-obvious.  I'm under the impression serializers are largely defined by just those two methods (to_representation and to_internal_value).

jchansen: If I understand what you did, you set up `to_internal_value` so that we could send PUT/PATCH requests that followed the form a) `{status_type=1}` or b) `{status_type: {id:1}}`, yes?  It might be worth considering (not needed for this PR) creating our own Serializer that captures that behavior - if that's the approach we want to follow, I'd like to apply it to all serializers.  I think most of the v2 serializers only allow the `{status_type=1}` format for now, but I also know that breaks the DRF forms.  Which I wasn't concerned about considered I Model.objects.None()'d all the querysets so the pages would load in a reasonable time.  Part of me wants to standardize on a single format, either a or b, and make that the standard.  Still debating in my mind, but leaning towards b.  I think I started using the PK serializer for v2 because it was the "easy choice" from what DRF supplied out of the box...

api/v2/serializers/details/quota_request.py
```diff
@@ -0,0 +1,161 @@
...
+            'status',
+            'created_by',
+            'user',
+            'identity',
+            'provider',
+            'admin_message',
+            'quota'
+        )
+
+class UserQuotaRequestSerializer(serializers.HyperlinkedModelSerializer):
```
jchansen: Are we "PEP8ing"?  If so, I think we need an extra line return before the `UserQuotaRequestSerializer` class definition.

api/v2/views/base.py
```diff
@@ -0,0 +1,153 @@
...
+            )
+            raise exceptions.ParseError(detail=message)
+        except IdentityMembership.DoesNotExist:
+            message = (
+                "The identity '%s' does not have a membership"
+                % identity_id
+            )
+            raise exceptions.ParseError(detail=message)
+
+    @unresolved_requests_only
```
jchansen: Nice.  I hadn't seen the decorators applied yet.  Can you apply multiple decorators to a method?

jmatt: Yep. Use functools.wraps() to ensure mutliple decorators work. They are evaluated top down. But if you don't use classes and wraps, it may not compose properly.

api/v2/views/base.py
```diff
@@ -0,0 +1,153 @@
...
+
+        if instance.is_denied():
+            self.deny_action(instance)
+
+    @unresolved_requests_only
+    def update(self, request, *args, **kwargs):
+        """
+        Update the request for the specific identifier
+        """
+        return super(BaseRequestViewSet, self).update(request, *args, **kwargs)
```
jchansen: Is this function calling itself?  It looks like it's defined in `BaseRequestViewSet` and then calls itself, and then calls itself again...?

core/email.py
```diff
@@ -175,6 +174,49 @@ def email_from_admin(username, subject, message, html=False):
                       cc=[email_address_str(from_name, from_email)],
                       html=html)
 
+def send_approved_quota_email(user, request, reason):
+    """
+    Notify the user the that their request has been approved.
+    """
+    template = """
```
jchansen: Is there a way to write the email text in `"""text"""` while still maintaining visual indentation?  This looks hideous.

steve-gregory: no.

core/models/group.py
```diff
@@ -182,9 +182,10 @@ def approve_quota(self, request_id):
         super(IdentityMembership, self).save()
         try:
             from service.tasks.admin import set_provider_quota,\
```
jchansen: (opinion) Would love to standardize on the parenthesis syntax for line-breaking imports over the backslash syntax.  The backslashes seem so...brutish.

core/email.py
jchansen: Waaaaaaah!!

erbriones: Hey @jchansen ,

I'm going to be working on adding email templates as a separate feature. This will move all these templates into actual template files which can be sent out.

api/v2/views/base.py
erbriones: @jchansen This is calling the parents implementation. The decorator is wrapping the request.

api/v2/serializers/details/quota_request.py
erbriones: @jchansen I Think your right that we could turn this into our own field serializer. I would say override constructor to take a model keyword argument which would be used for query. Could name it something like ModelIdRelatedField.

jmatt:
I was pretty involved in this from the outset. So I'm biased and obviously approve. But I appreciate the comments.

:boom:

PR #21, opened 2015-05-01
Email templates

erbriones:
**Add email template support**

### Features
- Add `send_email_template` function to directly send template emails.
- Create templates for existing emails.
- Update existing emails to use new templates.


steve-gregory:
Looks good to me!

jchansen:
Nice Job Evan!

[LGTM](http://lgtm.in/i/6vIElBjdq)

jmatt:
Looks good to me too. I'm going to merge so we all start using it immediately. I haven't tried all the email endpoints but seems to work. We should confirm this before the code move on Tuesday.

PR #22, opened 2015-05-02
nginx conf

erbriones:
**Atmosphere settings using nginx.**

### Features
- Includes static file and proxy settings for Atmosphere.
- Can auto generate nginx configuration using make.


jmatt:
Use nginx and uwsgi to serve Atmosphere.

Verified working by @erbriones and @jmatt. Early review by @steve-gregory.

:boom:

PR #23, opened 2015-05-04
Add v2 allocation request

erbriones:
**Add api v2 for submitting and updating allocations.**

### Features
- API Support for allocation requests
- Emails for new, approved and denied requests.
- Updates users allocation when approved

### DIFF
--- a/api/v2/serializers/details/__init__.py
+++ b/api/v2/serializers/details/__init__.py
@@ -1,4 +1,6 @@
 from .allocation import AllocationSerializer
+from .allocation_request import AllocationRequestSerializer,\
+    UserAllocationRequestSerializer
 from .identity import IdentitySerializer
 from .image import ImageSerializer
 from .image_tag import ImageTagSerializer
--- a/api/v2/serializers/details/allocation_request.py
+++ b/api/v2/serializers/details/allocation_request.py
@@ -1,0 +1,167 @@
+from rest_framework import exceptions, serializers
+from core.models import AllocationRequest, Allocation, Identity, AtmosphereUser as User
+from core.models.status_type import StatusType
+from api.v2.serializers.summaries import (
+    IdentitySummarySerializer,
+    UserSummarySerializer,
+    ProviderSummarySerializer,
+    AllocationSummarySerializer,
+    StatusTypeSummarySerializer
+)
+
+
+class UserRelatedField(serializers.PrimaryKeyRelatedField):
+
+    def get_queryset(self):
+        return User.objects.all()
+
+    def to_representation(self, value):
+        user = User.objects.get(pk=value.pk)
+        serializer = UserSummarySerializer(user, context=self.context)
+        return serializer.data
+
+
+class IdentityRelatedField(serializers.RelatedField):
+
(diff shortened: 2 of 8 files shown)


api/v2/serializers/details/__init__.py
```diff
@@ -1,4 +1,6 @@
 from .allocation import AllocationSerializer
+from .allocation_request import AllocationRequestSerializer,\
```
jchansen: Have we decided on an official convention for line-breaking imports yet?  I'm still a fan of using the () syntax to put each on a different line.

erbriones: I don't think we have decided on the convention. I've been sticking with \ because majority of the could still uses the \ style convention.

jchansen:
I don't expect the UI will be using this endpoint for a while, but I have no objection to it existing.  If we want to get users to request allocation and quota independently of each other (so I create one request for a quota modification and a second request to obtain more AUs) then we'll need to refactor the UI a fair amount to support that.

If that is indeed the direction we want to go (@edwins will need to weigh in, though likely in an out-of-chat design discussion), then I think refactoring phases should look like this

1. Modifying the existing quota_requests endpoint to allow an admin to also modify a users allocation.  Possibly rename to resources_requests since it encompasses both quota and allocation.

2. Refactor UI to provider an independent experience for quota and allocation requests.  Connect allocation experience to /allocation_requests endpoint.

3. Remove ability to set allocation from /quota_requests endpoint, rename back to quota_requests if was renamed to /resource_requests.

jchansen:
LGTM

[LGTM](http://lgtm.in/i/QWiMMPD7q)

erbriones:
I believe this was well discussed that they were separate end points and that we were planning on having separate requests for different resource types. In the current sense most users are requesting allocation resources and not a quota change. I've already opened a pull request for adding allocation request support and modified the existing resource request to be allocation specific. Additional, locations for requesting allocation need to be added to the UI. Quota is something that needs to be changed far less than allocation.

edwins:
I had a user experience discussion with Evan yesterday. Managing allocation and quota requests separately are ok for a first pass. However, moving forward users and operators consider this as one request and do not fully appreciate the nuance and frankly, arbitrary distinction that has been made between what was defined as a quota and allocation. These requests should be modeled as a unified request.

If it helps understand another perspective, my opinion is that time is just another resource that a user may want to increase, like cpu, storage, or ram. At some point, we'll also want to have a network bandwidth as well. 

If it makes more sense, should we unify the implementation of allocations and quota code as well?

erbriones:
We can merge the two requests into one request if that is what we want. Our initial decision was to create these as separate requests. Both v1 and v2 have been setup to support them as separate requests. If decide to merge them I would suggest dropping the request endpoint from v1.

steve-gregory:
Whats the current status on this PR? Are we planning on merging the feature in its current state or refactoring (in which case we should close this and continue conversation via JIRA)?

erbriones:
Closing request. It has been decided that Allocation and Quota will be merged into a resource requset.

PR #24, opened 2015-05-06
Add maintenance permissions

jmatt:
Add maintenance and permissions system to the api.

Different permissions can be found in api.permissions. These permissions are applied to views from which the all API inherits. Views can be found in api.views and api.v2.views.base modules.

PEP8 Compliance for all non-compliant code.


api/v2/views/tag.py
```diff
@@ -1,30 +1,23 @@
...
-
         serializer.save(user=self.request.user)
 
     def get_permissions(self):
-        method = self.request.method
-        if method == 'POST':
-            self.permission_classes = (IsAuthenticatedOrReadOnly,)
-
-        return super(viewsets.ModelViewSet, self).get_permissions()
+        if self.request.method in ["POST", "PUT", "DELETE"]:
```
jchansen: Nice, much simpler than what I wrote.

api/v2/views/identity.py
```diff
@@ -1,10 +1,11 @@
...
 from core.models import Identity, Group
-from api.v2.serializers.details import IdentitySerializer
 from core.query import only_current_provider
 
+from api.v2.serializers.details import IdentitySerializer
+from api.v2.views.base import AuthViewSet
+
 
-class IdentityViewSet(viewsets.ReadOnlyModelViewSet):
+class IdentityViewSet(AuthViewSet):
```
jchansen: This should be AuthReadOnlyViewSet?

api/v2/views/provider.py
```diff
@@ -1,27 +1,28 @@
...
-    permission_classes = (IsAuthenticated,)
     http_method_names = ['get', 'head', 'options', 'trace']
 
     def get_permissions(self):
         method = self.request.method
         if method == 'DELETE' or method == 'PUT':
-            self.permission_classes = (IsAdminUser,)
-
-        return super(viewsets.GenericViewSet, self).get_permissions()
+            self.permission_classes += (CloudAdminRequired,)
```
jchansen: Could be refactored similar to tags using self.request.method in ['DELETE', 'PUT]

api/v2/views/tag.py
```diff
@@ -1,30 +1,23 @@
...
 from core.models import Tag
+
+from api.permissions import CloudAdminRequired
 from api.v2.serializers.summaries import TagSummarySerializer
-from api.v2.permissions import IsAdminOrReadOnly
+from api.v2.views.base import AuthReadOnlyViewSet
 
 
-class TagViewSet(viewsets.ModelViewSet):
+class TagViewSet(AuthReadOnlyViewSet):
```
jchansen: This should be CloudAdminOrReadOnly?  Expected behavior:

Anyone can view tags.
Only authenticated users can create tags.
Only admins can modify or delete tags.

api/v2/views/tag.py
```diff
@@ -1,30 +1,23 @@
...
     """
     API endpoint that allows tags to be viewed or edited.
     """
+
     queryset = Tag.objects.all()
     serializer_class = TagSummarySerializer
-    permission_classes = (IsAdminOrReadOnly,)
     max_paginate_by = 1000
 
     def perform_create(self, serializer):
```
jchansen: Need to make sure tags with the same name can't be created.

erbriones:
Reviewed large subset of all the endpoints and maintenance permissions are being correctly applied.

jmatt:
Addressed @jchansen comments.

PR #29, opened 2015-06-03
Dd hotfix

steve-gregory:
These changes to be applied ASAP

### DIFF
--- a/atmosphere/celery_router.py
+++ b/atmosphere/celery_router.py
@@ -43,6 +43,8 @@ PERIODIC_TASKS = [
         "check_image_membership", "update_membership_for",
         "clear_empty_ips", "clear_empty_ips_for",
         "remove_empty_networks_for",
+        "reset_provider_allocation",
+        "monthly_allocation_reset"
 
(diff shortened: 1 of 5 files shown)


amercer1:
+1 Now that you have time on the first of the months, watch Game Of Thrones!

erbriones:
The commit a1353aec seems to be unrelated to the hotfix and looks like it needs more work (provider hardcoded). This commit should be pulled out from this pull request.

steve-gregory:
@erbriones Both commits are required for DD to be fixed, Unless you see some problem with the commit, it will stay in the pull request and should be evaluated as such.

-Steve

erbriones:
@steve-gregory The commit for the celery tasks is a feature/improvement. While the other commits relate directly to fixing the bug in sending emails. I see these as unrelated requests.

steve-gregory:
After reviewing with @erbriones, a 'new improvement' will be added to JIRA for extending functionality to the celery task to be used in a less-iplant-exclusive manner.

PR #31, opened 2015-06-23
Remaining changes from dinelli's

steve-gregory:
These changes were made in dinelli's and didn't get merged into flawless-falcon. I hand merged the conflicts, mostly related to v2 APIs moving away from __x__y__z__... in favor of core.query and the new periodic task to be executed monthly (Resetting provider's allocation for all non-staff users)

One *new* change is the addition of 'patch' as an acceptable http_method for Auth-protected endpoints.


erbriones:
I do notice some of these commits were cherry-picked back into upstream separately due to code paths being different between versions. Not sure about the others.

steve-gregory:
It appears in that example that those changes were blown away in a later fix found during the E-E or F-F cycle.. But this is an area where *dead code* can be introduced which isn't great.. I'll pull the instance you pointed to, and attempt to do a more thorough search of other dead code points.

steve-gregory:
Walked through the comment-chain and I'm pretty sure that things have been appropriately cleaned up during the E-E and F-F cycles (with the exception of permissions).. Code is currently deployed to 'mickey' and is running identically to F-F.

erbriones:
Is this necessary? We have a current stable master branch. If we individually cherry-pick the items not present we will get a much better merge. History is also going to loose some transparency with duplicate commits.

steve-gregory:
It has normally been standard process to "bring branches forward" as part of the release, mostly to avoid confusion when moving between branches.. Could you elaborate more on how this would be better dealt with in master?

erbriones:
So any specific commit you could git cherry-pick and select only the ones that aren't the same. Like all of my commits are already in the main branches. They just have different hash due to the cherry-pick not cleanly applying.

steve-gregory:
I took this chat offline and spoke with @erbriones regarding the PR and we have both determined that it would be worthwhile to merge the branch in, and to continue to monitor these types of problems and look for different solutions (when changing code in an 'old-branch' and, differently, in a 'new-branch') as we move forward.

PR #32, opened 2015-06-25
Refactor Allocation 'exception' message and - timedeltas

steve-gregory:
As it turns out, - timedelta's are quite confusing to 'humans'..

For example, humans understand: 1 hour, 30 minutes OVER your allotted allocation.
Humans do NOT understand: -1 days, 22 hours, 30 minutes OVER your allotted allocation.

To remedy this, we treat timedeltas as a 'positive-only' value, accompanied by a 'is_over (allocation)' boolean value to indicate whether that time should be considered "The amount OVER allocation" or "The amount UNDER allocation"


erbriones:
:+1 Looks to specifically fix the issue and updates the test helpers no glaring issues I can see.

PR #33, opened 2015-06-25
Image creator

steve-gregory:
This branch is 'synced up' with Troposphere's instance_launch_wizard branch, the PRs should be evaluated and merged together to avoid problems...
NOTE: There is still a LOT of discrepency between "Image Vs. Application", "version vs machine", "imageversion vs applicationversion", etc. going on... AFTER Troposphere and Atmosphere receive these PRs I have every intent of cleaning all of that up and changing the Tropo UI experience to provide a better experience, rather than "matching/parity" which was the intent of these two PRs.

----

Big Features:
* Moving from Application--> ProviderMachine to Application-->ApplicationVersion-->ProviderMachine
* 'update' function added to /v2/provider_machines
* Refactor of the Image and ProviderMachine serializers to accomidate the change above
* Created new Summary and Details serializers (And related field) For "ImageVersion"


steve-gregory:
Administrative FYI: I will probably squash all of these commits prior to accepting the PR..

core/models/version.py
```diff
@@ -0,0 +1,140 @@
...
+      - Installed Software
+      - Excluded Files
+    This is a container for that information.
+
+    NOTE: Using this as the 'model' for DB moving to ID==UUID format.
+    """
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    application = models.ForeignKey("Application", related_name="versions")
+    fork_version = models.ForeignKey("ApplicationVersion", blank=True, null=True)
+    name = models.CharField(max_length=32, blank=True, null=True)#Potentially goes unused..
```
steve-gregory: Think "1.0" here

core/query.py
```diff
@@ -71,3 +103,12 @@ def _active_identity_membership(user, now_time=None):
...
         member__user__username=user.username)
+
+def _query_membership_for_user(user):
+    """
+    All *Memberhsips use 'group' as the keyname, this will check
+    that the memberships returned are only those that the user is in.
+    """
+    if not user:
+        return None
+    return Q(group__id__in=user.group_set.values('id'))
```
steve-gregory: To be used in the "Rework" for *Memberships..

api/v1/serializers/machine_request_serializer.py
```diff
@@ -36,20 +36,20 @@ class MachineRequestSerializer(serializers.ModelSerializer):
                                          source='new_machine_owner',
                                          queryset=AtmosphereUser.objects.all()
     )
-    vis = serializers.CharField(source='new_machine_visibility')
-    version = serializers.CharField(source='new_machine_version',
+    vis = serializers.CharField(source='new_application_visibility')
```
erbriones: This should probably be renamed to `visible` or `visibility` rather than abbreviated.

steve-gregory: This is API v1, doing so would affect backwards compatibility.. However I agree on API v2 name changing!

api/v2/views/image.py
```diff
@@ -35,20 +34,33 @@ def get_queryset(self):
...
+            admin_list = Image.objects.filter(
+                only_current_apps(),
+                id__in=image_ids)
+            return (public_image_set | my_images | privately_shared | admin_list).distinct()
         else:
             return Image.objects.none()
 
     def retrieve(self, request, pk=None):
-        queryset = Image.objects.all()
+        queryset = self.get_queryset()
```
erbriones: This can be improved by calling self.get_object and returning the instance.

erbriones:
From what I can see looks good.

api/v1/serializers/machine_request_serializer.py
erbriones: Ah yes didn't notice that.

api/v2/views/image.py
steve-gregory: Reverted back to standard behavior, which I will assume already does the things you've mentioned above :)

steve-gregory:
Worked with @erbriones to verify that the code works as expected. Functional testing completed on this PR

PR #34, opened 2015-06-25
Ansible improve logging and errors

jmatt:
Improve Ansible logging and error handling. We now log all aggregate statistics about playbooks used during deploy and raise an exception if anything is unreachable (dark) or reports a failure.

See:

https://pods.iplantcollaborative.org/jira/browse/ATMO-884
https://pods.iplantcollaborative.org/jira/browse/ATMO-885


service/tasks/driver.py
```diff
@@ -907,6 +908,9 @@ def _deploy_init_to(driverCls, provider, identity, instance_id,
         playbooks = ansible_deploy_to(instance.ip, username, instance_id)
         _update_status_log(instance, "Ansible Finished for %s." % instance.ip)
         logger.debug("_deploy_init_to task finished at %s." % datetime.now())
+    except AnsibleDeployException as exc:
+        logger.exception(exc)
+        _deploy_init_to.retry(exc=exc)
```
steve-gregory: After talking to @amercer1 it seems like 'when Ansible fails there is usually a reason' and continued retries may be futile... Perhaps we should consider raising-out here?

steve-gregory:
:+1: Looks good to me

erbriones:
Straight forward implementation! :+1: Looks good takes care of reporting missing information and adding missing error handling.

service/tasks/driver.py
jmatt: There are still problems with network partitioning where it will eventually resolve. that's why we still retry some. But it's lowered to 10 retries and a max of 32 minutes.

PR #37, opened 2015-07-16
add multiple field lookup mixin

erbriones:
Add support for looking up objects by there id, uuid or by another field name.


steve-gregory:
:+1:

jmatt:
+1

PR #40, opened 2015-07-28
Implement Volume v2

erbriones:
* Implement  volume creation.
* Implement volume  updating. 
* Implement volume deletion with cascade deletion support.
* Update service exceptions to use new base exception `ServiceException`
* New `VolumeError` exception for volume errors.
* New `NotFound` exception for resources that are not found.

### DIFF
--- a/api/v2/serializers/details/__init__.py
+++ b/api/v2/serializers/details/__init__.py
@@ -10,12 +10,42 @@ from .instance import InstanceSerializer
 from .project import ProjectSerializer
 from .project_instance import ProjectInstanceSerializer
 from .project_volume import ProjectVolumeSerializer
-from .provider import ProviderSerializer, ProviderTypeSerializer, PlatformTypeSerializer
+from .provider import ProviderSerializer, ProviderTypeSerializer,\
+    PlatformTypeSerializer
 from .provider_machine import ProviderMachineSerializer
 from .quota import QuotaSerializer
-from .resource_request import ResourceRequestSerializer, UserResourceRequestSerializer
+from .resource_request import ResourceRequestSerializer,\
+    UserResourceRequestSerializer
 from .size import SizeSerializer
 from .status_type import StatusTypeSerializer
 from .tag import TagSerializer
 from .user import UserSerializer
-from .volume import VolumeSerializer
+from .volume import VolumeSerializer, UpdateVolumeSerializer
+
+__all__ = (
+    "AllocationSerializer",
+    "BootScriptSerializer",
+    "IdentitySerializer",
+    "ImageBookmarkSerializer",
+    "ImageSerializer",
+    "ImageTagSerializer",
+    "ImageVersionSerializer",
+    "InstanceSerializer",
+    "InstanceTagSerializer",
+    "PlatformTypeSerializer",
+    "ProjectInstanceSerializer",
+    "ProjectSerializer",
+    "ProjectVolumeSerializer",
+    "ProviderMachineSerializer",
+    "ProviderSerializer",
+    "ProviderTypeSerializer",
+    "QuotaSerializer",
+    "ResourceRequestSerializer",
+    "SizeSerializer",
+    "StatusTypeSerializer",
+    "TagSerializer",
+    "UpdateVolumeSerializer",
+    "UserResourceRequestSerializer",
+    "UserSerializer",
+    "VolumeSerializer"
+)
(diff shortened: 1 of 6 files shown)


erbriones:
Still need to check that we handle all error cases for `create` and `delete`.

api/v2/serializers/details/__init__.py
```diff
@@ -10,12 +10,42 @@
...
+from .resource_request import ResourceRequestSerializer,\
+    UserResourceRequestSerializer
 from .size import SizeSerializer
 from .status_type import StatusTypeSerializer
 from .tag import TagSerializer
 from .user import UserSerializer
-from .volume import VolumeSerializer
+from .volume import VolumeSerializer, UpdateVolumeSerializer
+
+__all__ = (
```
steve-gregory: Whats the goal of __all__ here? Is this a 'best practice' to be implemented else where?

service/volume.py
```diff
@@ -67,9 +103,47 @@ def create_volume(esh_driver, identity_uuid, name, size,
         metadata=metadata,
         snapshot=snapshot,
         image=image)
+
+    if not success and raise_exception:
+        raise exceptions.VolumeError("The volume failed to be created.")
```
steve-gregory: This error hides whether or not the failure was a result of "Us" (Atmosphere, bad process, etc.) or "Them" (Cloud failure, cloud error, etc.).. Perhaps we could split "ServiceException" into CloudService Vs. APIService and return a more appropriate error message + code based on that?

service/volume.py
```diff
@@ -8,10 +6,13 @@
...
 from service.cache import get_cached_driver
-from service.driver import _retrieve_source
+from service.driver import _retrieve_source, prepare_driver
 
+from service import exceptions
 from service.instance import boot_volume_instance
-from service.exceptions import OverQuotaError
+
+# FIXME: fix prepare_driver to take a user directly
+Request = namedtuple("request", ["user"])
```
steve-gregory: 'prepare_driver' was actually a derivative of https://github.com/iPlantCollaborativeOpenSource/atmosphere/blob/volume_v2/service/driver.py#L122-L138 -- Perhaps we should modify that function to match the behavior you are looking for, and do away with 'prepare_driver' altogether, since the WSGIRequest is inconsequential here

service/volume.py
```diff
@@ -67,9 +103,47 @@ def create_volume(esh_driver, identity_uuid, name, size,
...
+
+    # if cascade True and snapshots exist delete all snapshots
+    if cascade:
+        snapshots = esh_volume.list_snapshots()
+        for snapshot in snapshots:
+            driver.destroy_snapshot(snapshot)
+
+    # destroy the volume successfully or raise an exception
+    if not driver.destroy_volume(esh_volume):
+        raise Exception("Encountered an error destroying the volume.")
```
steve-gregory: Unsure if generic exception belongs here? Perhaps a CloudService/APIService exception instead to give users a little more guidance in the status code & message?

steve-gregory:
Nice job evan! I commented on a few minor details, but all and all this looks really good :) and it was one of the most complex API endpoints in V1 (Second only to instance), so we've definitely come a long way!

api/v2/serializers/details/__init__.py
erbriones: The purpose of `__init__.py` is for creating a package. The variable `__all__` restricts what is imported when importing the entire package. See https://docs.python.org/2/tutorial/modules.html#importing-from-a-package

The second reason is that we import everything and never use it which will fail flake8 test.

I decided to add it because it doesn't take too much extra work to keep the namespace from being polluted if everything was imported.

service/volume.py
erbriones: Ideally this exception should contain additional information and at least derive from `ServiceException` for now it is a blanket exception.

service/volume.py
erbriones: I could see `VolumeError` being used for actual errors and something like `NotFound` or a derivative of `ServiceException` being more of a warning or bad request.

service/volume.py
erbriones: I do like the name of `prepare_driver` I'm going to be working on moving Credentials into a permission class and authorize a persons credentials before creating a driver. These changes could simplify the driver code.

steve-gregory:
:+1: -- Refactor work could be done in a separate release, LG2M for a merge

PR #42, opened 2015-08-06
Image creator

steve-gregory:
These API changes are required in tandem with the 'image_creator' UI changes made on Troposphere (https://github.com/iPlantCollaborativeOpenSource/troposphere/pull/101)


erbriones:
From what I can see makes sense. Lets go ahead and merge it in.

PR #43, opened 2015-08-12
Conf templates

jmatt:
Configuration Templates
---------------------------------

Reimplement the .dist configuration system to use Jinja2 templates. This supports our goal to have reproducible, well defined deploys and environments. It will be used by our new Ansible deploy system, Clank.

**Implementation**

```
./variables.ini.dist
```
This is the only .dist file. It should contain all the variables required to configure Atmosphere with no values. This is only by convention. The script only looks at the .ini file.

```
./variables.ini
```
The source of all variables used in templates. Sections are only there for semantics, for now. So, variables are considered global. The variables are read using python's ConfigParser.

```
./scripts/generate_config.py
```
This script is for developers and is used internally by the nginx make. It tests templates, backs up and creates configs using Jinja2 templates. Use the _--help_ option for additional details.

```
./atmosphere/settings/local.py.j2
./atmosphere/settings/secrets.py.j2
./extras/apache/atmo.conf.j2
./extras/nginx/Makefile.j2
./extras/nginx/site.conf.j2
./extras/nginx/locations/atmo.conf.j2
./extras/nginx/locations/flower.conf.j2
./extras/nginx/locations/jenkins.conf.j2
./extras/nginx/locations/lb.conf.j2
```

Jinja2 configuration files. Typically the configuration file name + .j2. This is configurable in the _./scripts/generate_config.py_ script.


erbriones:
Gone ahead and tested and is working as expected. I'm going to go ahead and merge in.

PR #45, opened 2015-08-27
Add two more attributes to /v2/instance_history objects

steve-gregory:
Added 'image' and 'provider' to instance history to avoid deep-nested queries clientside. See https://github.com/iPlantCollaborativeOpenSource/troposphere/pull/126 for details on how this changed the UI implementation.


steve-gregory:
Merging after review with @erbriones  (and required changes were made)

PR #47, opened 2015-09-17
Include 'Monitor machines' as a periodic task

steve-gregory:
For the incana-incana release..

use 'prune_machines' to end date images that have been deleted out of band.
use 'monitor_machines' to 'pull in' images that have been created out of band, and update sharing (both on the DB and on the cloud) to reflect the user's privacy choices.


steve-gregory:
Tested end dating all of our oldest providers (by force -- since they are inactive) and running 'prune_machines' on provider 4,5,6. The results match up as expected. Cascading 'end dates' also work as expected.

PR #51, opened 2015-09-25
Add machine request views

prosif:
This PR adds a view/serializer for machine requests and implements V1 auto imaging approval in V2.


steve-gregory:
I worked with joseph to verify this works on chromogenic. Rest of the changes that follow LG2M!

PR #53, opened 2015-10-03
Don't allow attach/detach when instance is in a non-active state

steve-gregory:


steve-gregory:
Simple bugfix -- Going to merge this one

PR #57, opened 2015-10-26
Improve how configuration files are generated.

lenards:
## Background 

We have been using a script, `./scripts/generate_config.py`, to assist in creating a baseline configuration for settings within various environments (local dev, `atmo-dev`, `atmobeta`, and production). This existing does have some dependencies that the improved version will continue to use (``Jinja2``). 

Related to [tropo PR 180](https://github.com/iPlantCollaborativeOpenSource/troposphere/pull/180).

## What's been *improved* ... 

- no global namespace 
- more explicit about failure (in the spirit of #163)
- warning 
- verifying newly added variables 
- test with dry run 
- ANSI shell text colors 
- new name, location

### No Global namespace 

The previous version merged all _"sections"_ (`[nginx]`, `[apache]`, `[local.py]`) into a single, global namespace. This was unexpected, confusing, and required "prefixing" if you wanted different values for a variable, like `URL`. Since the config format has _sections_, you would example them to actually be *scoped*. 

*Note:* there is now a `[COMMON]` section to allow for "global" or _reused_ values that you'd like to have a _common_ value across sections. _However_, if you wish to override that variable within a *section*, you can. 

### Be more explicit about failure

As noted in #163, when we fail to generate the necessary configuration - we should be clear about that.  The most important failure communication is when the local environment is missing a *required* value. This improved script will halt and output the variables that are missing values. Overall, failures are now explicitly noted (even color coded, more on that later).  

### Warning 

We now indicate when there are unused variables within the `variables.ini` to help assist in managing various programming environments (these are even color coded).

### Verifying newly added variables 

There is a new ``--verify`` option for the script that will compare the variables present in the local ``variables.ini`` with those expected in ``variables.ini.dist``. This is just one way we to better communicate new variables added to support features. 

So doing the following will check the section defined between the 2 files: 
```bash
./configure --verify
```

### Test with ``--dry-run``

An argument to run the script without outputting the configuration files was added to help if troubleshooting your variables.ini. Simply add the flag when calling the script:
```bash
./configure --dry-run
```
*Note:* this will not output the configuration files nor backups of those configuration files. 

### ANSI shell text colors 

An optional dependency on `colorama` was added to provide ANSI shell text colors. This has been added to `dev_requirements`. However, the new script will seamless run without this Python module. The colors improve the communication, but we do not want the absence of this module to prevent running the script. 

### New name, location 

The name of the script is now `configure`. Since we've included a Python shebang in the file, we opted to remove the `.py` extension. This will allow for a (potentially) familiar `./configure && make` step for setting up an environment. 

### Examples 

Failed due to `variables.ini.dist` and `variables.ini` issues... 
screen shot 2015-10-26 at 2 39 38 pm

Successful dry run... 
screen shot 2015-10-26 at 2 42 26 pm

Missing or inactive `virtualenv`...
screen shot 2015-10-26 at 2 58 20 pm


lenards:
@steve-gregory 

I added the following to `variables.ini.dist` as 'shared', aka `COMMON`, variables: 

```ini
[COMMON]
SERVER_URL =
LDAP_SERVER =
LDAP_SERVER_DN =
TOKEN_EXPIRY_TIME_DAYS =
DJANGO_SERVER_URL =
```

We may want to eliminated `DJANGO_SERVER_URL` and change it to `SERVER_URL`. 

Also, the path to `atmosphere` is used in three sections - all variables have different names. 

We may want to consider merging `local.py` and `secret.py` sections in `variables.ini` and just altering the mapping of the sections in `variables_templates.json`. 

Thoughts?

lenards:
Need review from @steve-gregory - assigning...

steve-gregory:
LG2M :+1: See troposphere for some comments regarding my thoughts on how to print warnings. Thanks for putting in the work!

lenards:
I merged [Tropo::PR-180](https://github.com/iPlantCollaborativeOpenSource/troposphere/pull/180), do you want to do the honors here? 

:soon: :shipit:

PR #58, opened 2015-10-27
A call to is_valid on a serializer shouldn't have side effects.

cdosborn:
This change shouldn't have any negative effects on the code base. [Here](https://github.com/iPlantCollaborativeOpenSource/atmosphere/blob/bf8854654638a0e4473432087bf9d07475d1b65a/api/v1/views/profile.py#L45-L49) is the code that uses this modified serializer.


steve-gregory:
:+1: Overviewed w/ Connor. All looks good.

PR #59, opened 2015-10-28
Resolution for stop instances.

lenards:
We are unable to stop running instances for the latest version of the Atmo API ([ATMO-1054](https://pods.iplantcollaborative.org/jira/browse/ATMO-1054)).


steve-gregory:
:+1: Thank you sir.

PR #64, opened 2015-11-05
Provides an approach for rendering auth mock settings when in local development. 

lenards:
Subsequent runs of `./configure` will cause the local, mock authentication backend to be disabled (via a commented out block in [local.py](https://github.com/lenards/troposphere/blob/mock-auth-options-in-localpy-j2/troposphere/settings/local.py.j2#L78-L92)). This fosters a situation where individuals do not run the configuration generation script frequent which leads to local environments being out-of-sync with the configuration and variables on `HEAD`. 

`USE_MOCK_AUTH` will control if the mock authentication backend is enabled. 

For internal variable configuration management, set to the following:

```ini
[local.py]
...
MOCK_USER =
USE_MOCK_AUTH = False
```

*Note:* Setting as `USE_MOCK_AUTH = ` (so, equal to nothing) will also behave in beta, staging, and production as we desire (e.g. - the appropriate authentication will be used).


lenards:
@steve-gregory - can you take a close look here? 

This will require actions in our internal configuration management of variables (as you'd expect). 

- [ ] provide values for new variables introduced within _internal_ configuration management

steve-gregory:
:+1: Ready for merge!

PR #65, opened 2015-11-05
Allow Administrators to end-date Users access to an Identity

steve-gregory:
To be reviewed concurrently with the Troposphere PR of the same name..

~~This PR adds `UUID` Fields to all available V2 models. Join tables (project_instance, project_volume, etc.) were excluded because they are "un-managed tables" (AKA: Django-problems)~~

This PR also adds an 'end_date' field to IdentityMembership and updates queries throughout the V2 API to ensure that, if the user has access to an Identity, that their IdentityMembership's end_date is "valid" (Not null, OR end date has not yet happened yet).

Finally, added a V2 endpoint for /identity_memberships


api/v1/views/instance.py
```diff
@@ -99,6 +99,9 @@ def get(self, request, provider_uuid, identity_uuid):
         if not esh_driver:
             return invalid_creds(provider_uuid, identity_uuid)
         identity = Identity.objects.get(uuid=identity_uuid)
+        # Probably redundant
+        if not user.can_use_identity(identity.id):
+            return invalid_creds(provider_uuid, identity_uuid)
```
lenards: Does `not esh_driver` catch this case as well? I'm curious why it might be redundant.

api/v2/serializers/details/provider.py
```diff
@@ -25,12 +25,19 @@ class Meta:
...
+    type = ProviderTypeSerializer()
+    virtualization = PlatformTypeSerializer()
+    sizes = SizeSummarySerializer(source='size_set', many=True)
+    is_admin = serializers.SerializerMethodField()
+
+    def get_is_admin(self, provider):
+        user = self.context['request'].user
+        if user.is_staff or user.is_superuser:
+            return True
+        return False
```
lenards: we could just do: 
```python
...
return user.is_staff or user.is_superuser
```

No need for using the conditional to create a Boolean to return (or am I missing a subtlety here?) :8ball:

core/migrations/0037_add_end_date_identitymembership.py
```diff
@@ -0,0 +1,24 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
```
lenards: :heart_decoration: `__future__`

api/v2/serializers/details/provider.py
steve-gregory: Mildly-intentional use of this pattern because it allows us to drop in more 'If X: return True" conditionals.. But I have nothing against changing it.

core/models/user.py
```diff
@@ -22,15 +21,26 @@ def provider_ids(self):
...
+    def can_use_identity(self, identity_id):
+        return self.current_identities.filter(id=identity_id).count() > 0
+
     def select_identity(self):
         """
         Set, save and return an active selected_identity for the user.
         """
         # Return previously selected identity
-        if self.selected_identity and self.selected_identity.is_active():
+        if self.selected_identity and self.selected_identity.is_active(self):
```
lenards: Hey - we talked about this - I think the keyword argument helps make this _more_ understood (bad English :us: > :gb:)

```python
self.selected_identity.is_active(user=self)
```

api/v2/serializers/details/provider.py
lenards: Okay. I'm okay with it. Was just curious if there was a reason.

core/models/user.py
lenards: I think this is the only change I would like to see before merging.

steve-gregory: Agreed! Will fix now :+1:

PR #66, opened 2015-11-12
Include basic account bootstrapping for newly authenticated users.

steve-gregory:
Create a new setting to control whether or not the first identity should be auto-created for a user that has been deemed authorized into Atmosphere. This is a stop-gap measure, and this feature will be fleshed out at a later point in time to include which provider should/should not be used, what quota and allocation should be assigned to new users, etc.


lenards:
Looks good :mag_right: 
:shipit:

lenards:
@steve-gregory - leaving the branch ...

PR #67, opened 2015-11-12
Avoided shallowing exception & opt for unicode output strings for enhanced output

lenards:
Currently, the automation used via ansible has the limitation of an ASCII charset and the output from our enhanced scripts cause encoding failures. 

This removes the offending characters from the configure script.


steve-gregory:
:shipit:

PR #68, opened 2015-11-13
Altered template for nginx Makefile; correct copy-paste errors.

lenards:
This template for the extra nginx makefile has the old-style `generate_configs.py` for various targets. 

It also contained a copy-paste error for generating the nginx-jenkins and nginx load balancer.


steve-gregory:
Lg2m, good catch

PR #70, opened 2015-11-19
Small fixes

cdosborn:
The flower broker url is not what flowers like :tulip:. See @steve-gregory for a good reason why it breaks celery/flower.

The second small fix is for an incorrect error message.


steve-gregory:
:shipit:

PR #71, opened 2015-11-19
Fix calls to feedback/support

cdosborn:
This fixes the 404 that results from clicking feedback/support at the bottom.

This needs to be cherry-picked into master.


steve-gregory:
:shipit: && I will trickle-down merge to master.

PR #73, opened 2015-11-20
Respond with 409 and inform user when a Provider is Inactive

steve-gregory:
[JIRA] (https://pods.iplantcollaborative.org/jira/browse/ATMO-782)

When a provider is marked inactive, let the user know via a non-200 API status code with a descriptive error message: `Cannot create driver on an inactive provider: iPlant Workshop Cloud - Tucson`

Providers are 'inactive'  when they have reached their `end_date` or when the `is_active` flag is toggled off.

### DIFF
--- a/api/email.py
+++ b/api/email.py
@@ -21,7 +21,7 @@ from service.volume import create_volume
 from service.exceptions import OverQuotaError
 
 from api.serializers import VolumeSerializer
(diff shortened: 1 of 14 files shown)


steve-gregory:
Will Self-:shipit: tomorrow.

PR #74, opened 2015-11-23
End date resource requests

prosif:
Allow users to close their own resource requests.

This is done by end dating resource requests when a PATCH is made to the resource_requests endpoint with a status of "closed." 

An end dated request also allows a user to make another request without their old request needing to be fully deleted from the database.

### DIFF
--- a/api/v2/views/base.py
+++ b/api/v2/views/base.py
@@ -178,7 +178,7 @@ class BaseRequestViewSet(MultipleFieldLookup, AuthViewSet):
                 if self.request._method == "PATCH":
                     instance = serializer.save(status=StatusType.objects.get(id=serializer.initial_data['status']))
                 else:
-                    instance = serializer.save(end_date=timezone.now())
+                    instance = serializer.save()
 
             if instance.is_approved():
                 self.approve_action(instance)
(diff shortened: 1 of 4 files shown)


lenards:
@steve-gregory - I'm just curious about this PR since it relates to Troposphere as well.

api/v2/views/base.py
```diff
@@ -175,11 +175,17 @@ def perform_update(self, serializer):
                 instance = serializer.save(end_date=timezone.now(),
                                            membership=membership)
             else:
-                instance = serializer.save(end_date=timezone.now())
+                if self.request._method == "PATCH":
+                    instance = serializer.save(status=StatusType.objects.get(id=serializer.initial_data['status']))
+                else:
+                    instance = serializer.save(end_date=timezone.now())
```
steve-gregory: I don't know if we should be saving the 'end_date' every time here. That is not the correct behavior (Although i see above thats what was done previously).. @prosif Can you make a note in JIRA to address this in a different commit/PR?

steve-gregory: Actually, i see below you did put in end_date appropriately. Drop the 'end_date=' kwarg from this line.

steve-gregory:
LG2M. :shipit:

PR #78, opened 2015-12-02
fix typo in exempt allocation list

prosif:
The exempt allocation list incorrectly contained allocations with threshold of -1, which was an empty list. This has been corrected to delta = -1, which will fix monthly allocation reset happening for users who should be exempt.


steve-gregory:
owned. I feel so :shipit:

PR #80, opened 2015-12-10
Improved the status output of the `atmosphere` service script.

lenards:
The script now reports the status of other (included/dependent) daemons in the same order that is used for `start|stop|restart`. 

Also, an indication between the uWSGI processes for atmosphere & troposphere is included (along with clarifying output for nginx, celery, redis, etc). 

In addition to that, a `status` method for `celerybeat` is now included from the generic init script example within the celery [repo](https://github.com/celery/celery/blob/master/extra/generic-init.d/celerybeat).

Tested and verified locally on vagrant. 

Targeting `master` (aka Larping Loon) release.


lenards:
Related back to [ATMO-1043](https://pods.iplantcollaborative.org/jira/browse/ATMO-1043).

api/v2/views/machine_request.py
```diff
@@ -34,9 +67,9 @@ def perform_create(self, serializer):
                 status=status,
                 old_status="processing",
```
steve-gregory: This line considered harmful (?)

steve-gregory:
uWSGI part LG2M. Not so sure about machine request.

api/v2/views/machine_request.py
lenards: Yes, still harmful. I'm not introducing that line here.

steve-gregory: Ok LG2M!

steve-gregory:
:shipit:

lenards:
Thanks.

PR #82, opened 2015-12-15
Fix identitymembership-detail hyperlink relationship error

lenards:
Within the `beta` & `dev` environments, we are seeing errors regarding a _missing_ hyperlink relationship:
screen shot 2015-12-15 at 2 14 07 pm

It appears this [_detail_](https://github.com/iPlantCollaborativeOpenSource/atmosphere/blob/master/api/v2/serializers/details/identity_membership.py#L21) serializer was intended to be added in the future - but has not. 

I found this inclusion of the serializer for the `url` field within a local branch on Steve's development machine. I was able to reproduce the above error within my vagrant API server and verified this fix there as well.


lenards:
Was going to revert the PR but opted for a new PR, #83.

PR #87, opened 2016-01-06
Raise v1 machine exception

prosif:
Addresses ATMO-976
Return invalid_provider_identity when an identity doesn't exist on a provider. Previously returned an HTML response to a serializer that caused DRF issue.


steve-gregory:
:shipit:

PR #88, opened 2016-01-08
Fix forms in api v1

cdosborn:


cdosborn:
@steve-gregory Are you okay with the merge?

cdosborn:
Sometimes you just gotta do, watcha gotta do

PR #89, opened 2016-01-13
Pull in rtwo that fixes setting provider quota

cdosborn:
**Note:** this will need to be picked into kicking-kestrel


cdosborn:
screen shot 2016-01-13 at 4 35 00 pm

cdosborn:
Cherried here 0dd7be

PR #90, opened 2016-01-19
Provided functionality from #84 as a patch to kicking-kestrel

lenards:
This brings over 3 commits from #84.


steve-gregory:
:shipit: Thank you @lenards :100:

PR #92, opened 2016-01-26
Fix API v2 /volume POST and DELETE methods

steve-gregory:
The perform_create and perform_delete methods, previously, were pointing to bad functions and variable names. As well, some variables were being requested that were not necessary for CREATE. All of these errors have been fixed. Description is now a passable kwarg when POSTing to volume.

An update to rtwo (0.2.18) was required to move forward.

I did test creating volumes with image_id and was unable to perform the action on OpenStack Havana. We can revisit why Images aren't being copied into Volumes when we are running a later version of OStack.

### DIFF
--- a/api/v2/views/volume.py
+++ b/api/v2/views/volume.py
@@ -86,7 +86,6 @@ class VolumeViewSet(MultipleFieldLookup, AuthViewSet):
         instance_source = data.get("instance_source")
         identity = instance_source.get("created_by_identity")
         provider = identity.provider
-
         try:
             esh_volume = create_volume_or_fail(name, size, self.request.user,
                                                provider, identity,
--- a/api/v2/views/volume.py
+++ b/api/v2/views/volume.py
@@ -104,7 +104,8 @@ class VolumeViewSet(MultipleFieldLookup, AuthViewSet):
         except VOLUME_EXCEPTIONS as e:
             raise exceptions.ParseError(detail=e.message)
         except Exception as exc:
-            logger.exception("Error occurred creating a v2 volume")
+            logger.exception("Error occurred creating a v2 volume -- User:%s"
+                             % self.request.user)
             return Response(exc.message, status=status.HTTP_409_CONFLICT)
 
     def perform_destroy(self, instance):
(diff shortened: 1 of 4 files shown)


service/volume.py
```diff
@@ -57,22 +53,20 @@ def _update_volume_metadata(esh_driver, esh_volume,
 
 
 def restrict_size_by_image(size, image):
-    image_size = image._connection.get_size(image._image)
+    image_bytes = image._image.extra['image_size']
```
lenards: Can you always depend on `_image` and `_image.extra` not being equal to `None`?

api/v2/views/volume.py
```diff
@@ -104,6 +103,9 @@ def perform_create(self, serializer):
             return inactive_provider(pna)
         except VOLUME_EXCEPTIONS as e:
             raise exceptions.ParseError(detail=e.message)
+        except Exception as exc:
+            logger.exception("Error occurred creating a v2 volume")
```
lenards: Since we're starting to send logs to `lumberjack`, should be include some information that might track the `Exception` back to an instance-alias, volume-alias, or something that would get us back to a person having a problem?

lenards:
Okay, I confirmed that `rtwo==0.2.18` was pushed to [pypi](https://pypi.python.org/pypi/rtwo/0.2.18). 

It looks good. I think it can go :shipit:. 

I do think we want to consider if more logging would help trace problems back to instances/volumes/etc within OpenStack.

service/volume.py
steve-gregory: In this case, yes:
```
In [5]: image = driver.get_machine("12345678-1234-1234-4321-123456789098")
type(image)
In [6]: type(image)
Out[6]: rtwo.machine.OSMachine

In [7]: type(image._image)
Out[7]: libcloud.compute.base.NodeImage

In [8]: type(image._image.extra)
Out[8]: dict
```

api/v2/views/volume.py
steve-gregory: Added 'user' *only* because:
1. Its guaranteed to be available && will help track down specific users who can complain.
2. volume alias isn't available here (Because create failed)

lenards:
You have *rocked* the code! :shipit:

steve-gregory:
Thanks for the review @lenards !

PR #93, opened 2016-01-27
Fix periodic task 'prune_machines'

steve-gregory:
Fix for ATMO-1114

Fixed prune machines so that it will prune invalid Apps, Versions and ProviderMachines based on cloud *and* database values.

Added some convenience methods for `end_date_all()` and `_current_versions()`

### DIFF
--- a/service/tasks/monitoring.py
+++ b/service/tasks/monitoring.py
@@ -92,14 +92,13 @@ def prune_machines_for(provider_id, print_logs=False, dry_run=False, forced_remo
     NOTE: BEFORE CALLING THIS TASK you should ensure that the AccountProvider can see ALL images. Failure to do so will result in any unseen image to be prematurely end-dated and removed from the API/UI.
     """
     provider = Provider.objects.get(id=provider_id)
+    now = timezone.now()
     if print_logs:
         import logging
         import sys
         consolehandler = logging.StreamHandler(sys.stdout)
         consolehandler.setLevel(logging.DEBUG)
         celery_logger.addHandler(consolehandler)
-    mach_count = ver_count = app_count = 0
-    now = timezone.now()
     celery_logger.info("Starting prune_machines for Provider %s @ %s"
                        % (provider, now))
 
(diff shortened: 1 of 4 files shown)


service/tasks/monitoring.py
```diff
@@ -118,8 +121,41 @@ def prune_machines_for(provider_id, print_logs=False, dry_run=False, forced_remo
...
+            Q(num_machines__gt=0) &
+            Q(versions__machines__instance_source__end_date__isnull=True)
+        ),
+        # AND application has already been end-dated.
+        end_date__isnull=False
+    )
+
+    for app in improperly_enddated_apps:
+        app.end_date_all(now)
+        app_count += 1
```
lenards: this loop appears 3 times - can we could something like: 

```python
def _perform_end_date(apps, end_dated_at):
     count = 0
     for app in apps:
          app.end_date_all(end_dated_at)
          count += 1
     return count
```

I know it's a bit of a quibble - but this is a monstrously long function.

service/tasks/monitoring.py
```diff
@@ -118,8 +121,41 @@ def prune_machines_for(provider_id, print_logs=False, dry_run=False, forced_remo
...
+        # Contains at least one version without an end-date OR
+        (Q(num_versions__gt=0) & Q(versions__end_date__isnull=True)) |
+        # conatins at least one machine without an end-date
+        (
+            Q(num_machines__gt=0) &
+            Q(versions__machines__instance_source__end_date__isnull=True)
+        ),
+        # AND application has already been end-dated.
+        end_date__isnull=False
+    )
```
lenards: If worth extracting this query into a function? It might not be. But `prune_machine_for` has varying levels of operational abstraction ... or could be abstracted (might be a better way to put it).

lenards:
We can move forward with this. But, applying some "single level of abstraction" to `prune_machines_for` would simplify understanding the steps within this _task_.

lenards:
If priority on [_ATMO-1114_](https://pods.iplantcollaborative.org/jira/browse/ATMO-1114) was higher than minor - I would say "go ahead a ship" :shipit: - but this could be a good reference for later (perhaps).

service/tasks/monitoring.py
steve-gregory: Agreed. will refactor this

service/tasks/monitoring.py
steve-gregory: :+1: Will add to refactor list

steve-gregory:
Changes left to be made:

- [ ] Split up the prune_machine method into smaller pieces
- [ ] Split up the complex queries used in prune_machine

steve-gregory:
@lenards to you for final review

lenards:
This is *ready* - :shipit: 

Good work @steve-gregory.

steve-gregory:
@lenards thanks for the tips && review! I definitely think the code is more readable now than it would have been 21 days ago. :+1:  Merging..

PR #94, opened 2016-01-27
Update image version

prosif:
Addresses ATMO-1138

Allow users to update minimum requirements of the images they own.
Users can update either min_cpu or min_mem or both, and both will default to 0 if no value is given.
CPU is currently limited to positive integers less than 16, and memory is limited to positive integers less than 32 * 1024 (32 GB).

##### Troposphere PR #256 is dependent on this PR

### DIFF
--- a/service/instance.py
+++ b/service/instance.py
@@ -1188,8 +1188,11 @@ def check_application_threshold(
     application = Application.objects.filter(
         versions__machines__instance_source__identifier=boot_source.identifier,
         versions__machines__instance_source__provider=core_identity.provider).distinct().get()
-    threshold = application.latest_version.get_threshold()
-#TODO: This is *NOT* good logic. Instead, look at 'boot_source' and determine what ApplicationVersion (If any) it belongs to. THEN use that version to 'get_threshold().
+    try:
+        threshold = boot_source.current_source.application_version.get_threshold()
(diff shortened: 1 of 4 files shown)


service/instance.py
```diff
@@ -1188,7 +1188,11 @@ def check_application_threshold(
...
         versions__machines__instance_source__identifier=boot_source.identifier,
         versions__machines__instance_source__provider=core_identity.provider).distinct().get()
-    threshold = application.get_threshold()
+    try:
+        threshold = boot_source.current_source.application_version.get_threshold()
+    except:
+        threshold = application.latest_version.get_threshold()
+
     if not threshold:
         return
```
steve-gregory: The logic here should be:
try to retrieve threshold. If it fails -- There is no minimum threshold -- return.

steve-gregory:
LG2M! :shipit:

steve-gregory:
On a side note -- I see a lot of MB <==> GB conversions... Since we are doing 'minimum requirements' and given the sizes/flavors in most clouds, maybe its best to use GB *exclusively*? Seems silly to specify a minimum requirement of '512 MB' or '1.5Gb' (Although it is conceivable).. This would be an easy change to both PRs but I am curious what others think.

PR #96, opened 2016-01-28
Fix instance deleted

prosif:
Addresses ATMO-1141
When attempting to delete an instance that no longer exists, raise 404 error with descriptive error. 

Previously, there were checks in all instance actions **except** the delete action to see if the instance exists before continuing.

This PR adds that check before deleting an instance, and fixes an error in the exception that was being raised in all other actions in the case of a nonexistent instance. 

Here is a timeline of the original error returned by the API when deleting a nonexistent instance to the current error being returned.
errors


steve-gregory:
:shipit:

PR #97, opened 2016-01-29
Update user boot scripts deployment to include injection of $ATMO_ variables

steve-gregory:
This first PR include support for a single variable, $ATMO_USER that represents the user of the identity that created the instance.
Any other variables injected should also carry the $ATMO_ namespace.

Scripts that execute after this should be expected to "source the file $HOME/.bashrc for the deploying user (Usually root)".

A later version would ideally have ansible support. Will consult with @amercer1 about development on this front.

This will be patched into arturo to help a user get their image up and running properly.


steve-gregory:
Scratch -- Can't be merged into kicking as it was pre-merged into larping loon. Will update.

PR #98, opened 2016-02-01
Update user boot scripts deployment to include injection of $ATMO_ variables

steve-gregory:
This first PR include support for a single variable, $ATMO_USER that represents the user of the identity that created the instance.
Any other variables injected should also carry the $ATMO_ namespace.

Scripts that execute after this should be expected to "source the file $HOME/.bashrc for the deploying user (Usually root)".

A later version would ideally have ansible support. Will consult with @amercer1 about development on this front.

This will be patched into production to help a user get their image up and running properly.

### DIFF
--- a/core/models/boot_script.py
+++ b/core/models/boot_script.py
@@ -61,8 +61,8 @@ class BootScript(models.Model):
         return slugify(self.title).replace('-', '_')
 
     def get_text(self):
-        if self.script_type.name == 'Full Text':
-            return self.script_text
+        if self.script_type.name == 'Raw Text':
+            return self.script_text.strip ()  # Remove whitespace
         elif self.script_type.name == 'URL':
             return self._text_from_url()
 
--- a/core/models/boot_script.py
+++ b/core/models/boot_script.py
@@ -62,7 +62,7 @@ class BootScript(models.Model):
 
     def get_text(self):
         if self.script_type.name == 'Raw Text':
-            return self.script_text.strip ()  # Remove whitespace
+            return self.script_text.strip()  # Remove whitespace
         elif self.script_type.name == 'URL':
             return self._text_from_url()
 
--- a/service/deploy.py
+++ b/service/deploy.py
@@ -6,6 +6,7 @@ import os
 import sys
 import time
 
+from django.utils.text import slugify
 from django.utils.timezone import datetime
 
 from libcloud.compute.deployment import Deployment, ScriptDeployment,\
@@ -464,9 +465,27 @@ def wrap_script(script_text, script_name):
     * Chmod the file
     * Execute and redirect output to stdout/stderr to logfile.
     """
-    logfile = "/var/log/atmo/post_boot_scripts.log"
+    # logfile = "/var/log/atmo/post_boot_scripts.log"
     # kludge: weirdness without the str cast...
     script_text = str(script_text)
-    full_script_name = "./deploy_boot_script_%s.sh"
+    full_script_name = "./deploy_boot_script_%s.sh" % (slugify(script_name),)
     return ScriptDeployment(
         script_text, name=full_script_name)
+
+
+def _inject_env_script(username):
+    """
+    This is the 'raw script' that will be used to prepare the environment.
+    TODO: Find a better home for this. Probably use ansible for this.
+    """
+    return \
+        """#!/bin/bash -x
+atmo_user="%s"
+env_file="%s"
+env_vars=`cat $env_file`
+if [[ $env_vars == *"ATMO_USER="* ]]; then
+    sed -i "s/ATMO_USER=.*/ATMO_USER=\"$atmo_user\"/" $env_file
+else
+    echo "ATMO_USER=\"$atmo_user\"" >> $env_file
+fi
+""" % (username, "$HOME/.bashrc")
--- a/service/deploy.py
+++ b/service/deploy.py
@@ -6,6 +6,8 @@ import os
 import sys
 import time
 
+from django.template import Context
+from django.template.loader import render_to_string
 from django.utils.text import slugify
 from django.utils.timezone import datetime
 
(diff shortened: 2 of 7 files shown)


core/models/boot_script.py
```diff
@@ -61,8 +61,8 @@ def get_title_slug(self):
         return slugify(self.title).replace('-', '_')
 
     def get_text(self):
-        if self.script_type.name == 'Full Text':
-            return self.script_text
+        if self.script_type.name == 'Raw Text':
+            return self.script_text.strip ()  # Remove whitespace
```
lenards: `whitespace == ` :japanese_goblin:

service/deploy.py
```diff
@@ -464,9 +465,27 @@ def wrap_script(script_text, script_name):
...
+        """#!/bin/bash -x
+atmo_user="%s"
+env_file="%s"
+env_vars=`cat $env_file`
+if [[ $env_vars == *"ATMO_USER="* ]]; then
+    sed -i "s/ATMO_USER=.*/ATMO_USER=\"$atmo_user\"/" $env_file
+else
+    echo "ATMO_USER=\"$atmo_user\"" >> $env_file
+fi
+""" % (username, "$HOME/.bashrc")
```
lenards: I agree with the `TODO` here - we might want to make this a template, similarly to how we handle rendering into the [email feedback](https://github.com/iPlantCollaborativeOpenSource/atmosphere/tree/master/core/templates/core/email), etc. 

Maybe `core/templates/scripts`...

service/tasks/driver.py
```diff
@@ -32,7 +32,7 @@
 from core.models.identity import Identity
 from core.models.profile import UserProfile
 
-from service.deploy import check_process, wrap_script, echo_test_script,\
+from service.deploy import _inject_env_script, check_process, wrap_script, echo_test_script,\
```
lenards: Maybe I'm missing the mark on the purpose of `_method_name` - but when I see an underbar method imported - that's a signal for me to pay extra attention. I believe that with `_` preview we're saying that `_inject_env_script` is _intended_ to be use *within* `service.deploy` - but not outside. 

I know we need to forge ahead ... or [_ford the river_](http://coldfrontmag.com/wp-content/uploads/2013/06/2-Trail_069.jpg). But just something I _note_ as I look at code. 

Let me know if I'm off-base.

service/tasks/driver.py
```diff
@@ -557,9 +563,9 @@ def get_remove_status_chain(driverCls, provider, identity, instance):
     return start_chain
 
 
-def get_chain_from_build(driverCls, provider, identity, instance,
-                         username=None, password=None, redeploy=False,
-                         deploy=True):
+def get_chain_from_build(
+        driverCls, provider, identity, instance, core_identity,
+        username=None, password=None, redeploy=False, deploy=True):
```
lenards: :+1: on the re-format.

service/tasks/driver.py
```diff
@@ -589,9 +595,9 @@ def print_chain(start_task, idx=0):
         mystr += print_chain(task, idx+1)
     return mystr
 
-def get_chain_from_active_no_ip(driverCls, provider, identity, instance,
-                                username=None, password=None, redeploy=False,
-                                deploy=True):
+def get_chain_from_active_no_ip(
+        driverCls, provider, identity, instance, core_identity,
+        username=None, password=None, redeploy=False, deploy=True):
```
lenards: :+1:

service/tasks/driver.py
```diff
@@ -613,16 +619,16 @@ def get_chain_from_active_no_ip(driverCls, provider, identity, instance,
...
     end_chain.link(deploy_start)
     return start_chain
 
 
-def get_chain_from_active_with_ip(driverCls, provider, identity, instance,
-                                  username=None, password=None,
-                                  redeploy=False, deploy=True):
+def get_chain_from_active_with_ip(
+        driverCls, provider, identity, instance, core_identity,
+        username=None, password=None, redeploy=False, deploy=True):
```
lenards: :+

service/tasks/driver.py
```diff
@@ -847,6 +867,9 @@ def destroy_instance(instance_alias, user, core_identity_uuid):
       max_retries=10)
 def deploy_script(driverCls, provider, identity, instance_id,
                   script, **celery_task_args):
+    """
+    #TODO: Should this be ansible now?
+    """
```
lenards: This is a question I _had_ too. I'm not sure about the answer. But it seems like a reasonable division of labor.

lenards:
Overall - I think you are _good to go_ here. 

I think we want to consider: 
- [ ] finding a home for `_inject_env_script` in a more fitting place (maybe template?)
- [ ] determine how this _could_ be handled by `atmosphere-ansible`.

lenards:
But, :shipit:

lenards:
I see this is still "Open" - but I'm okay with getting this in now. :ship:

service/tasks/driver.py
lenards: If we _intend_ for `_inject_env_script` to be used outside of the defining file, I would change this to `inject_env_script`.

core/models/boot_script.py
steve-gregory: The irony here...

service/tasks/driver.py
steve-gregory: Noted for fixing

steve-gregory:
Updated the essentials. Will leave TODO's in the PR so that they can be addressed by @amercer1 at a later date.

lenards:
I think this is ready. You got the :shipit: from me.

lenards:
Bon voyage :ship:

steve-gregory:
Will merge into prod tonight.

lenards:
Okay.

PR #102, opened 2016-02-10
Respect admin_url parameter for resource admin emails

prosif:
If an `admin_url` parameter is passed in a POST to create a resource request, change the admin URL emailed to administrators to respect it. This is done by taking the base admin URL given by the client, and appending the ID of the newly created request to it.

If an `admin_url` parameter does not exist, continue sending the current Django admin link.


steve-gregory:
:shipit: LG2M. Thanks @prosif

PR #103, opened 2016-02-15
Add close action to machine request

prosif:
Add function called when `request.is_closed()` is true. Currently has no functionality, but prevents an error from being thrown when an imaging request is denied.


steve-gregory:
Nice work! :shipit:

PR #104, opened 2016-02-16
Default private access list

prosif:
If an image is private, share the image with the user and admins by default. Also remove duplicate usernames from access list.


steve-gregory:
:shipit: Thanks @prosif

PR #105, opened 2016-02-16
Admin/super users can act on closed requests

prosif:
Allow privileged users to act on closed requests

### DIFF
--- a/api/v2/views/base.py
+++ b/api/v2/views/base.py
@@ -7,7 +7,7 @@ from rest_framework.response import Response
 from rest_framework.viewsets import ModelViewSet, ReadOnlyModelViewSet
 
 from core import exceptions as core_exceptions
-from core.models import IdentityMembership
+from core.models import IdentityMembership, CloudAdministrator
 from core.models.status_type import StatusType
 
 from api.permissions import (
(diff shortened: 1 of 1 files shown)


api/v2/views/base.py
```diff
@@ -24,8 +24,9 @@ def unresolved_requests_only(fn):
     @wraps(fn)
     def wrapper(self, request, *args, **kwargs):
         instance = self.get_object()
+        user_can_act = request.user.is_staff or request.user.is_superuser or CloudAdministrator.objects.filter(user=request.user.id)
```
steve-gregory: This queryset should includ `.count()` to ensure you get the truthy/falsy-ness you are looking for, and the Speed of an SQL lookup (Without having to generate an entire set of CA objects that you won't use!)

I will fix this one before merging in..

steve-gregory:
Having reviewed Troposphere, I'm going to mark this: :shipit:

api/v2/views/base.py
erbriones: :8ball: https://docs.djangoproject.com/en/1.9/ref/models/querysets/#django.db.models.query.QuerySet.exists

lenards: that examples seems add to me: 
```python
entry = Entry.objects.get(pk=123)
if some_queryset.filter(pk=entry.pk).exists():
    print("Entry contained in queryset")
```

Couldn't you just do: 
```python 
if some_queryset.filter(pk=123).exists():
    entry = Entry.objects.get(pk=123)
    # do work on `entry`
    print("Entry contained in queryset")
```

Maybe it's just the scenario in the example lacks _full_ context.

steve-gregory: @lenards you can't use .get because a "lookup miss" will generate a ModelDoesNotExist error.

So you are either dealing with `.filter()` logic 'in an if statement' or `.get()` logic inside 'try/except' statement..

lenards: _/puzzle_ ... if `.get()` throws a `ModelDoesNotError`, then isn't the Django QuerySet example likely to generate an exception?

erbriones: Correct! That example probably has an error in it. What it probably should show is.
```python
some_queryset = Entry.objects.filter(pk=123)
if some_queryset.exists():
    entry = some_query.first()
    print("Entry contained in queryset")```

Or it should show the example you listed above.

PR #106, opened 2016-02-17
Jetstream machine request fixes

prosif:
Hot new feature: Add to the request's access list if it exists instead of completely ignoring it.
Also add user groups from that access list to `new_version_membership`


steve-gregory:
Ready for merge. :shipit: !

PR #107, opened 2016-02-22
Backport latest Jetstream changes to LL

steve-gregory:
This PR contains all the changes made to Jetstream that had not made it back to the 'main stream branches'.

notable changes listed and tested:
- Create `openrc` Credentials for a given account.
- Move from `globus` to `CAS+OAuth`
- Create (at least one -- but allow multiple) accounts when new user has no identity to select.
- Switches between LDAP and Django as the E-Mail Authority
- Switches between 'Old' v2.0 auth calls and 'New' v3 auth calls

- [x] Verify that latest `requirements.txt` (Notably, rtwo && chromogenic) work :100: on mickey.
- [x] Verify that rtwo can create/attach volumes using "The cinder methods" in latest requirements update.

Will seek review once the final two checks are marked.


steve-gregory:
This PR ready for final review && merge.

core/email.py
```diff
@@ -171,6 +222,7 @@ def email_to_admin(
     if not username and not user_email:
         username, user_email = sendto, sendto_email
     elif not user_email:  # Username provided
+        # TODO: Pass only strings, avoid passing 'User' object here.
```
lenards: Agree! [P.O.L.A.](http://c2.com/cgi/wiki?PrincipleOfLeastAuthority) _(plus some [yelling about the topic](https://www.youtube.com/watch?v=eL5o4PFuxTY))_

scripts/add_new_accounts.py
```diff
@@ -11,7 +11,7 @@
 from core.models import Provider, Identity
 
 from service.accounts.openstack_manager import AccountDriver as OSAccountDriver
-
+from threepio import logger
```
lenards: Long Live C3PO - human cyborg relations!

service/accounts/openstack_manager.py
```diff
@@ -24,17 +30,31 @@
...
-class AccountDriver(CachedAccountDriver):
+class AccountDriver(BaseAccountDriver):
     user_manager = None
     image_manager = None
     network_manager = None
     core_provider = None
 
     MASTER_RULES_LIST = [
-        ("ICMP", 0, 255),
+        ("ICMP", -1, -1),
```
lenards: I have no idea what `-1` means here. I assume that someone would know that this is the Internet Control Message Protocol value?

lenards: Expanding the rest of the rules, no worries about the above comment.

service/accounts/openstack_manager.py
```diff
@@ -201,8 +240,19 @@ def init_security_group(self, username, password, project_name,
...
+        # user = user_matches[0]
+        # -- User:Keystone rev.
+        kwargs = {}
+        if self.identity_version > 2:
+            kwargs.update({'domain': 'default'})
+        user_matches = [u for u in self.user_manager.keystone.users.list(**kwargs) if u.name == username]
+        if not user_matches or len(user_matches) > 1:
+            raise Exception("User maps to *MORE* than one account on openstack default domain! Ask a programmer for help here!")
+        user = user_matches[0]
+        kwargs = {}
```
lenards: Do we _mean_ to reset `kwargs` here? Does project expect `'domain_id'` and not `'domain'`?

service/accounts/openstack_manager.py
```diff
@@ -253,8 +303,12 @@ def get_or_create_keypair(self, username, password, project_name,
...
         """
         clients = self.get_openstack_clients(username, password, project_name)
-        nova = clients["nova"]
-        keypairs = nova.keypairs.list()
+        if self.identity_version == 2:
+            nova = clients["nova"]
+            keypairs = nova.keypairs.list()
+        else:
+            osdk = clients["openstack_sdk"]
+            keypairs = [kp for kp in osdk.compute.keypairs()]
```
lenards: Okay - this is the _"either or"_ pattern you mentioned to me earlier. Got it :+1:

service/accounts/openstack_manager.py
steve-gregory: Yes. For now both keystone projects and users expect different versions of kwargs. I could explicitly rename them as list_kwargs, and maybe separate them..?

service/accounts/openstack_manager.py
```diff
@@ -273,12 +327,33 @@ def create_keypair(self, username, password,
...
+            keypair = osdk.compute.create_keypair(
+                name=keyname,
+                public_key=public_key)
         return keypair
 
+    def accept_shared_image(self, glance_image, project_name):
+        """
+        This is only required when sharing using 'the v2 api' on glance.
+        """
+        #NOTE: Abusing the 'project_name' == 'username' mapping. Future me will fix this.
```
lenards: Use `TODO:` or `FIXME:` - this is more than a _note_.

service/accounts/openstack_manager.py
```diff
@@ -516,22 +570,121 @@ def list_all_images(self, **kwargs):
...
-        return self.user_manager.get_project(project_name)
+    def get_project(self, project_name, **kwargs):
+        if self.identity_version > 2:
+            kwargs = self._parse_domain_kwargs(kwargs)
+        return self.user_manager.get_project(project_name, **kwargs)
 
     def _make_tenant_id_map(self):
         all_projects = self.list_projects()
         tenant_id_map = {project.id: project.name for project in all_projects}
         return tenant_id_map
```
lenards: Add space before `create_trust`.

service/accounts/openstack_manager.py
```diff
@@ -516,22 +570,121 @@ def list_all_images(self, **kwargs):
...
+    def get_user(self, user_name_or_id, **list_kwargs):
+        if self.identity_version > 2:
+            list_kwargs = self._parse_domain_kwargs(list_kwargs)
+        user_list = self.list_users(**list_kwargs)
+        found_users = [user for user in user_list if user.id == user_name_or_id or user.name == user_name_or_id]
+        if not found_users:
+            return None
+        if len(found_users) > 1:
+            raise Exception("User name/id %s matched more than one value -- Fix the code" % (user_name_or_id,))
+        return found_users[0]
```
lenards: :japanese_goblin:

service/deploy.py
```diff
@@ -158,10 +158,22 @@ def configure_ansible(logger):
 
 
 def build_host_name(ip):
+    """
+    Return the host name
+    * iPlant or raw
```
lenards: this is only returning "iplant" - in what case would `raw_hostname` get returned?

steve-gregory: If we ever lost DNS and couldn't keep `vmX-Y` as a value? Probably never.. really keeping them both around to make swapping back to jetstream easier (if necessary)

service/machine.py
```diff
@@ -195,12 +195,26 @@ def share_with_admins(private_userlist, provider_uuid):
...
+    img = accounts.get_image(new_machine.identifier)
+    if hasattr(img, 'visibility'):  # Treated as an obj.
+        is_public = img.visibility == 'public'
+    elif hasattr(img, 'items'):  # Treated as a dict.
+        is_public = img.get('visibility','N/A') == 'public'
+
+    if is_public:
+        print "Marking image %s private" % img.id
+        accounts.image_manager.update_image(img, visibility='private')
+
```
lenards: I'm confused about how this is _private_ when the conditional is testing for it being public. I'm sure there is some contextual knowledge I lack.

service/accounts/openstack_manager.py
steve-gregory: From: http://docs.openstack.org/user-guide/cli_nova_configure_access_security_for_instances.html
```
To allow pinging of the instances, choose one of the following options:

Allow pinging from all IP addresses, specified as IP subnet 0.0.0.0/0 in CIDR notation.

$ nova secgroup-add-rule SECURITY_GROUP_NAME icmp -1 -1 0.0.0.0/0
This allows access to all codes and all types of ICMP traffic.
```

service/tasks/driver.py
```diff
@@ -971,7 +971,7 @@ def deploy_ready_test(driverCls, provider, identity, instance_id,
         "deploy_ready_test task %s/%s started at %s." %
         (current_count, total, datetime.now()))
     try:
-        # Check if instance still exists
+        # Sanity checks -- get your ducks in a row.
```
lenards: :duck:

service/machine.py
steve-gregory: Context is in the docs in the method you dont see:
```
    ASSERT: The image in 'new_machine' SHOULD BE private
    (Based on values in machine_request)
```

service/tasks/driver.py
steve-gregory: :duck: in a row

service/tasks/driver.py
```diff
@@ -980,13 +980,53 @@ def deploy_ready_test(driverCls, provider, identity, instance_id,
...
+    try:
+        kwargs = _generate_ssh_kwargs()
+        kwargs.update({'ssh_username': 'cirros', 'deploy': echo_test})
+        driver.deploy_to(instance, **kwargs)
+        celery_logger.debug(
+            "deploy_ready_test task %s/%s finished at %s." %
+            (current_count, total, datetime.now()))
+        return 'cirros'
+    except (BaseException, Exception) as exc:
+        pass
```
lenards: Extract method! Attempts 1, 2, 3 are exactly the same... 
```python
def connection_attempt(ssh_username, fail=False):
    try:
        kwargs = _generate_ssh_kwargs()
        kwargs.update({'ssh_username': ssh_username, 'deploy': echo_test})
        driver.deploy_to(instance, **kwargs)
        celery_logger.debug(
            "deploy_ready_test task %s/%s finished at %s." %
            (current_count, total, datetime.now()))
        return ssh_username
    except (BaseException, Exception) as exc:
        if(fail):
            # exception logging...
        else:
            pass
```

Then ... 
```python 
for ssh_username in ['ubuntu', 'centos', 'cirros']:
     connection_attempt(ssh_username)
connection_attempt('root', fail=True)
```

lenards: Or something like that :laughing:

service/accounts/openstack_manager.py
lenards: Maybe just a comment above second `kwargs` 

```
# reset before <blah call>
```

service/deploy.py
lenards: Okay. Just wasn't sure if that method body was missing an if-expression or something.

service/machine.py
lenards: Okay. That doesn't fully help. Really, a `new_machine` _should be_ private - by default. That's the expectation. But according to the code, `is_public` is True by default. We then _have to_ repair that and set it to private so the ASSERTION handles.

lenards:
:shipit:

lenards:
Merging by Steve's request.

PR #109, opened 2016-02-25
Add filtering to instance status history

prosif:
We currently treat "instance status history" as an instance launch history in Troposphere, displaying the list as a list of the instances you've launched.

This list is actually a list of all instance statuses, meaning one instance has multiple entries in the list, which goes against how we've been using it.

This PR allows for filtering `unique=True` to get a list with unique instance IDs, effectively giving us a list of the instances a user has launched.

### DIFF
--- a/api/v2/views/instance_history.py
+++ b/api/v2/views/instance_history.py
@@ -37,7 +37,7 @@ class InstanceStatusHistoryViewSet(MultipleFieldLookup, AuthReadOnlyViewSet):
     queryset = InstanceStatusHistory.objects.all()
     serializer_class = InstanceStatusHistorySerializer
     ordering = ('-instance__start_date', 'instance__id')
-    ordering_fields = ('-instance__start_date', 'instance__id')
+    ordering_fields = ('start_date', 'instance__id')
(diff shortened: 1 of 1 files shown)


api/v2/views/instance_history.py
```diff
@@ -36,8 +36,8 @@ class InstanceStatusHistoryViewSet(MultipleFieldLookup, AuthReadOnlyViewSet):
     """
     queryset = InstanceStatusHistory.objects.all()
     serializer_class = InstanceStatusHistorySerializer
-    ordering = ('-start_date', 'instance__id')
-    ordering_fields = ('start_date', 'instance__id')
+    ordering = ('-instance__start_date', 'instance__id')
+    ordering_fields = ('-instance__start_date', 'instance__id')
```
steve-gregory: I *think* you can leave `start_date` in the list for `ordering_fields` and still get what you want here...
Additionally, if one wanted to sort history by start date (on a given instance, for example) they should be able to accomplish that via `?ordering=start_date`

steve-gregory:
:shipit:

PR #111, opened 2016-03-02
Include new endpoint `/api/status/celery`

steve-gregory:
Endpoint was created at the request of QA, who would like to determine if celery is running prior to executing API tests.

If celery is running, the API returns:
```
HTTP 200 OK
Allow: GET, HEAD, OPTIONS
Content-Type: application/json
Vary: Accept

{
    "status": 200,
    "message": "Celery is running"
}
```

If celery is NOT running, the API returns:
```
HTTP 404 Not Found
Allow: GET, HEAD, OPTIONS
Content-Type: application/json
Vary: Accept

{
    "status": 404,
    "message": "Celery is NOT running"
}
```

### DIFF
--- a/api/status/urls.py
+++ b/api/status/urls.py
@@ -1,0 +1,14 @@
+# -*- coding: utf-8 -*-
+"""
+Routes for api status endpoints
+"""
+from django.conf.urls import patterns, include, url
+from rest_framework import routers
+
+from api.status import views
+
+router = routers.DefaultRouter(trailing_slash=False)
+router.register(r'celery', views.CeleryViewSet, base_name='celery')
+
+api_status_urls = router.urls
+urlpatterns = patterns('', url(r'^', include(api_status_urls)),)
--- a/api/urls.py
+++ b/api/urls.py
@@ -9,4 +9,6 @@ urlpatterns = patterns(
     '',
     url(r'', include("api.v2.urls", namespace="default")),
     url(r'^v1/', include("api.v1.urls", namespace="v1")),
-    url(r'^v2/', include("api.v2.urls", namespace="v2")))
+    url(r'^v2/', include("api.v2.urls", namespace="v2")),
+    url(r'^status/', include("api.status.urls", namespace="status"))
+    )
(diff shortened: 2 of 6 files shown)


api/status/urls.py
```diff
@@ -0,0 +1,14 @@
...
+from django.conf.urls import patterns, include, url
+from rest_framework import routers
+
+from api.status import views
+
+router = routers.DefaultRouter(trailing_slash=False)
+router.register(r'celery', views.CeleryViewSet, base_name='celery')
+
+api_status_urls = router.urls
+urlpatterns = patterns('', url(r'^', include(api_status_urls)),)
```
lenards: By the way - `patterns(...)` is being deprecated in Django 1.9. I believe the suggestion is for this to be a simple Python list or `url( , )`:
```python
urlpatterns = [
   url(r'^', include(api_status_urls))
] 
```
source: [troposphere/urls.py](https://github.com/iPlantCollaborativeOpenSource/troposphere/pull/302/files)

api/urls.py
```diff
@@ -9,4 +9,6 @@
     '',
     url(r'', include("api.v2.urls", namespace="default")),
     url(r'^v1/', include("api.v1.urls", namespace="v1")),
-    url(r'^v2/', include("api.v2.urls", namespace="v2")))
+    url(r'^v2/', include("api.v2.urls", namespace="v2")),
+    url(r'^status/', include("api.status.urls", namespace="status"))
+    )
```
lenards: Again, Django 1.10 is going to remove `patterns( ... )` ([info](https://github.com/lenards/troposphere/commit/1e9c284ad6a875a2129b2f1b64dcfda2bd5df7ae))

lenards:
I discussed with Steve that this won't indicate when the Celery processing is running, but for whatever reason not processing work-items (aka _tasks_). We both acknowledge that is a shortcoming of this - but having this simple indicator available externally is better than not having it (I'm on-board with that). 

:+1: 

:shipit:

steve-gregory:
I have noted your comments in JIRA for a future-feature :+1:

PR #112, opened 2016-03-03
Remove all old and unused scripts

steve-gregory:
Scripts that weren't removed had `django.setup()` line moved above atmosphere imports.


steve-gregory:
Self-:shipit:

PR #113, opened 2016-03-03
Image IDs on replicated clouds should be placed in the same version as a parent

steve-gregory:
The Old implementation would create a new version for each (provider, image_id) pairing, even if the image_id was an *exact match* to an already-existing version.

As a result, we ended up with lots of "Version 1.0.0" and "Version 1.0.0.0" that contained ProviderMachines that should have been in "Version 1.0".

This change will stop this problem moving forward. A one-off-fix will be required on the Production DB to 'clean' all versions that map to 1.0.0 or 1.0.0.0 that already have a matching PM. (Separate from this PR.)


steve-gregory:
Additional changes not related to 'Adding fuzzy search' prepared `monitor_machines_for` to run on the latest set of requriements for `mystifying-merlin` which no longer uses `image.is_public` as an attribute.

steve-gregory:
Self-:shipit:

PR #115, opened 2016-03-03
Filter image version machines

prosif:
Filter displayed machines an image version is available on, based on whether or not the user has access to the cloud the provider machine exists on.

This is done for the ProviderMachine viewset itself, and the nested `machines` field within an image version.


steve-gregory:
Great Job! :shipit:

PR #116, opened 2016-03-08
Use configure to take care of divergence between Master-stream and Jet-stream

steve-gregory:
The highlight reel:

- [x] Remove all references to "iPlant Cloud - Tucson"
- [x] Reference functions by 'functionName' in settings
- [x] Create a 'Default domain name'
- [x] Attempt to create instance moving to progressively higher size. If instance can not be launched at any size, validation fails.
- [x] Values for Role name, Email lookup function, IP lookup function
- [x] Update remaining references in glance between 'is_public'-->'visibility'
- [x] Move from settings to an EmailTemplate model and API that allows dynamic updating of values to e-mail.

Relevant JIRA:
[ATMO-1220 - ](https://pods.iplantcollaborative.org/jira/browse/ATMO-1220)
NOTE: Migrations are included in this change!
NOTE 2: Final checklist item added..Ready for merge!


steve-gregory:
@lenards This is ready for a final review. However, I would recommend holding off merge until you have Troposphere ready.. Then we can make final changes before sending it down to Jenkins!

configure
```diff
@@ -229,8 +229,13 @@ def _coerce(val):
...
     try:
-        return ast.literal_eval(val) if not val.isdigit() else val
+        if ('[' in val or '{' in val) and '__' not in val:
+            eval_fn = eval
+        else:
+            eval_fn = ast.literal_eval
+        return eval_fn(val) if not val.isdigit() else val
     except Exception as e:
+        #print "Exception occurred during evaluation of:%s - %s" % (val, e)
         return val
```
lenards: I'm curious what values of `val` are causing the need to do cause for that? 

I assume that `{` is for `{{ template_values}}`` - but I'm curious what we're doing now that uncovered this or what changed.

steve-gregory: This change away from literal_eval was required *soley* to bring in the change for `timedelta`... :+1:

steve-gregory: Also the `__` protection was recommended in many (How do i make eval safe) threads in S/O..

core/templates/core/email/imaging_request.html
```diff
@@ -1,9 +1,14 @@
...
+{% if support_email_header %}
+{{ support_email_header|safe }}
+{% endif %}
 {%if approved%}
 Your image request has been approved! The imaging process will
 begin shortly.
 {%else%}
 Your image request has been received. Upon staff approval, we will send you an e-mail to let you know that the the imaging process will begin shortly.
 {%endif%}
+{% endspaceless %}
```
lenards: `spaceless` for the win!

service/deploy.py
```diff
@@ -198,15 +196,49 @@ def configure_ansible(logger):
...
+    list_of_subnet = split_ip_address(ip)
+    return "%s%s%s%s.%s" % (
+        prefix, list_of_subnet[2], separator, list_of_subnet[3],
+        "jetstream-cloud.org")
 
 
 def iplant_hostname(ip):
-    list_of_subnet = ip.split(".")
-    return "vm%s-%s" % (list_of_subnet[2], list_of_subnet[3])
+    prefix = "js-"
```
lenards: Shouldn't this remain `vm`?

service/deploy.py
```diff
@@ -198,15 +196,49 @@ def configure_ansible(logger):
...
+    """
+    INPUT: IP Address (str in form: VVV.XXX.YYY.ZZZ)
+    js-YYY-ZZZ.jetstream-cloud.org
+    """
+    prefix = "js-"
+    separator = "-"
+    list_of_subnet = split_ip_address(ip)
+    return "%s%s%s%s.%s" % (
+        prefix, list_of_subnet[2], separator, list_of_subnet[3],
+        "jetstream-cloud.org")
```
lenards: Is there any way we might be able to make this a configuration-driven strategy?

I'm thinking of how DRF does things like paging arguments or something along those lines...

service/deploy.py
steve-gregory: Yep! typo!

steve-gregory:
@lenards it turns out you were right! I used a different configuration-driven strategy by allowing users to pass their own 'format string'

configure
lenards: :+1:

lenards:
Sweet! :+1:

service/deploy.py
```diff
@@ -196,49 +196,45 @@ def configure_ansible(logger):
...
-        "jetstream-cloud.org")
-
-
-def iplant_hostname(ip):
-    prefix = "js-"
-    separator = "-"
-    list_of_subnet = split_ip_address(ip)
-    return "%s%s%s%s.%s"\
-        % (prefix, list_of_subnet[2], separator, list_of_subnet[3],
-           "iplantcollaborative.org")
```
lenards: Happy to see installation-specific names get deleted! :tada:

lenards:
:shipit: :ex

atmosphere/settings/__init__.py
```diff
@@ -39,9 +39,7 @@
 
 # NOTE: first admin will be sender of atmo emails.
 ADMINS = (
-    ('Atmosphere Admin', 'atmo@iplantcollaborative.org'),
-    ('Steven Gregory', 'esteve@iplantcollaborative.org'),
-    ('Atmosphere Alerts', 'atmo-alerts@iplantcollaborative.org'),
+    ('AT LEAST ONE ADMIN REQUIRED', 'sends-email@if-debug-false.com'),
```
lenards: :+1:

PR #118, opened 2016-03-15
Fix: users see each others graphs

cdosborn:
The cache lookup did not take into account the instance uuid, when fetching
the instance's datapoints. The cache would return the latest datapoints for a
similar request made to any instance.

My reaction: (. _.) -> (0  .  @)


steve-gregory:
:shipit:

PR #119, opened 2016-03-15
Fix cache error, where users see incorrect graphs

cdosborn:
There was a hasty commit to fix this cache error. I reverted it for historical reference and added the proper fix.


cdosborn:
- [ ] Cherry into mystifying merlin

steve-gregory:
:shipit:

PR #123, opened 2016-03-23
Add __unicode__ for all core models

steve-gregory:
The `__unicode__` method makes it easier to determine information about the objects you are looking at in Django. A lot of core models were missing this method, in those instances I chose the most useful information to display per resource.

### DIFF
--- a/api/renderers.py
+++ b/api/renderers.py
@@ -25,9 +25,15 @@ class BrowsableAPIRenderer(renderers.BrowsableAPIRenderer):
     """
     'Custom' Browsable API Renderer
 
-    By setting 'display_edit_forms' to False in the context,
-    you can display the API without forms.
+    By returning an empty rendered HTML form
+    you can display the API quickly, without
+    having to deal with "select" queryset slow-downs.
     """
+
+    def get_rendered_html_form(self, data, view, method, request):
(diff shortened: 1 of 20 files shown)


api/renderers.py
```diff
@@ -25,10 +25,16 @@ class BrowsableAPIRenderer(renderers.BrowsableAPIRenderer):
...
-    By setting 'display_edit_forms' to False in the context,
-    you can display the API without forms.
+    By returning an empty rendered HTML form
+    you can display the API quickly, without
+    having to deal with "select" queryset slow-downs.
     """
 
+    def get_rendered_html_form(self, data, view, method, request):
+        if method in ['PUT','POST']:
+            return ""
```
steve-gregory: This single change resulted in (Easily) a 50x speed-up of the Browsable API.

JSON rendering was never affected.

steve-gregory:
Self-:shipit:

lenards:
:+1:

erbriones:
If you decorate the class with the `@python_2_unicode_compatible` you can future proof and use`__str__` over `__unicode__`.
https://docs.djangoproject.com/en/1.9/ref/utils/#django.utils.encoding.python_2_unicode_compatible

PR #124, opened 2016-03-28
Split instance activity

prosif:
Add "activity" field to instance serializers and InstanceStatusHistory model.
"Status" currently has the same functionality as before to maintain compatibility with the old UI.
"Activity" simply contains the text in the "status" field after the first dash.


core/models/instance.py
```diff
@@ -269,15 +271,26 @@ def update_history(
             last_history = InstanceStatusHistory.create_history(
                 status_name, self, size, self.start_date)
```
steve-gregory: activity was added in L244 but not here?

core/models/instance.py
```diff
@@ -223,23 +223,25 @@ def get_last_history(self):
                 name='Unknown Size', alias='N/A', provider=self.provider,
                 cpu=-1, mem=-1, root=-1, disk=-1)
             last_history = self._build_first_history(
-                'Unknown', unknown_size, self.start_date, self.end_date, True)
+                'Unknown', 'Unknown', unknown_size, self.start_date, self.end_date, True)
```
steve-gregory: Activity `None` here is fine.

steve-gregory:
Overall it looks good, a few minor changes should be made before merging.

core/models/instance.py
```diff
@@ -223,23 +223,25 @@ def get_last_history(self):
...
-            logger.info("First Update Unknown - Status name on instance %s: %s" % (self.provider_alias, status_name))
+            logger.info("First Update Unknown - Status name on instance \
+                        %s: %s - %s"
+                        % (self.provider_alias, status_name))
             # Instance state is 'unknown' from start of instance until now
             # NOTE: This is needed to prevent over-charging accounts
             status_name = 'unknown'
         first_history = InstanceStatusHistory.create_history(
-            status_name, self, size, start_date, end_date)
+            status_name, activity, self, size, start_date, end_date)
```
steve-gregory: Argument order seems odd here?

core/models/instance.py
```diff
@@ -589,14 +621,14 @@ def transaction(cls, status_name, instance, size,
                 "another transaction.")
 
     @classmethod
-    def create_history(cls, status_name, instance, size,
+    def create_history(cls, status_name, activity, instance, size,
```
steve-gregory: Maybe re-organize so its
`instance, size, status_name, activity=None, start_date=None, end_date=None):`

core/models/instance.py
steve-gregory: Isn't this line now invalid? @prosif

core/models/instance.py
```diff
@@ -590,13 +623,13 @@ def transaction(cls, status_name, instance, size,
 
     @classmethod
     def create_history(cls, status_name, instance, size,
-                       start_date=None, end_date=None):
+                       start_date=None, end_date=None, activity=None):
```
steve-gregory: Argument order is different from that in transaction and all `.create_history` above are using arg-order `status_name, activity, instance, size`..

steve-gregory:
- [X] Reorder migrations after latest commits merged ahead of this one.

PR #125, opened 2016-03-31
Add volume serializer start date

prosif:
Addresses ATMO-1116
Add `start_date` and `status` fields into VolumeUpdateSerializer, so that the response contains more relevant information.
Example in Troposphere
Before:
zibfg0sm13

After:
esyc6snpqd
This happens because backbone will take the response from the API and set its model to have the same attributes as the response. Since the response only contains partial information, the backbone model only contains a name and description until the volume is fetched again and the attributes are updated.


steve-gregory:
I think there is an alternative approach we should use.. Let's chat tomorrow

PR #126, opened 2016-03-31
Add machine request filtering

prosif:
If `?active=true` when retrieving machine requests, only display requests that are currently pending or have been created in the past 7 days.

### DIFF
--- a/api/v2/views/machine_request.py
+++ b/api/v2/views/machine_request.py
@@ -26,7 +26,7 @@ class MachineRequestViewSet(BaseRequestViewSet):
     model = MachineRequest
     serializer_class = UserMachineRequestSerializer
     admin_serializer_class = MachineRequestSerializer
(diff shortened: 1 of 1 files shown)


api/v2/views/machine_request.py
```diff
@@ -27,6 +30,20 @@ class MachineRequestViewSet(BaseRequestViewSet):
     ordering_fields = ('start_date', 'end_date', 'new_machine_owner__username')
     ordering = ('-start_date',)
 
+    def get_queryset(self):
+        if 'active' in self.request.query_params:
+            all_active = MachineRequest.objects.filter(
+                (
+                    Q(status__name='pending') |
```
steve-gregory: Can we change this to: `status__name_ne='closed'`? this way we can avoid 'losing track of' tickets that are open and in a state other than pending.

PR #129, opened 2016-04-06
Instance Actions Endpoint

steve-gregory:
- [X] Properly responds to GET calls on /v2/instances/<id>/action
- [X] New /v2/instance_actions endpoint
- [X] Form properly renders 'valid' data for the actions endpoint on BrowsableAPI
- [X] New file: api/decorators.py (Remove content from __init__.py)
- [X] New file: core/models/manager.py (Remove content from instance.py)
- [X] PEP8, Replace 'one-off' querys with more standardized `core.query`


core/query.py
```diff
@@ -47,6 +47,34 @@ def only_current_provider(now_time=None):
...
+        AND if it has not been end-dated before -now-
+        """
+        return (Q(source__end_date__isnull=True) |
+                Q(source__end_date__gt=now_time)) &\
+            Q(source__start_date__lt=now_time)
+
+    if not now_time:
+        now_time = timezone.now()
+    #NOTE: Purposefully absent: 'source_in_range'
+    return only_current() & _active_source()
```
steve-gregory: Wanted to mention that this change is being made, potentially avoids endusers "Losing scope" of an instance..

steve-gregory:
After further discussion with @lenards there is more work to be done here.

PR #131, opened 2016-04-11
Fixed incorrect attribute name problem in allocation.models.inputs.Ma…

julianpistorius:
…chine#__unicode__: `self.instance_source.identifier` should be `self.identifier`


julianpistorius:
Testing

PR #132, opened 2016-04-11
Remove 'Name:' which gets interpretted by RT as command

cdosborn:
Just a simple template change. RT tries to interpret the `Name:` header as a command.
When this gets merged, I'll cherry into NN.


julianpistorius:
Can one of the admins verify this patch?

steve-gregory:
@julianpistorius thanks for the ping!
:shipit:

PR #137, opened 2016-04-28
Add configuration for distributed routers per provider

steve-gregory:
This PR helps support the requirement for managing the even distribution of routers as new accounts are created for Atmosphere.

By default, the "list of routers" is set to one, and was originally stored in the "router_name" ProviderCredential.

To add a list of routers, create a new ProviderCredential:
`ProviderCredential.objects.get_or_create(provider=provider_object, key='public_routers', value='public_router1,public_router2,public_router3')`

To distribute the routers for the first time, run this command:
`./scripts/admin_redistribute_routers.py --provider 4`
And all users will have one of the three routers listed above.

To revert, remove the credential (Or update the value to contain `public_router` only)
and redistribute the routers:
`./scripts/admin_redistribute_routers.py --provider 4 --redistribute`


lenards:
This might be silly - but is there any benefit to making the `value` an actually valid syntactical Python list?

Something like:
```python 
ProviderCredential.objects.get_or_create(provider=provider_object, key='public_routers', value='[public_router1,public_router2,public_router3]')
```

I may just not have enough context to know the idioms of Django ORM - so this might be totally normal. I know I'd be fine with `value=public_router1,public_router2,public_router3`, myself, as I can easily use `value.split(',')` to get a list of values (so putting `[...]` around it might just be more hassle).

core/models/identity.py
```diff
@@ -157,6 +157,9 @@ def create_identity(cls, username, provider_location,
             c_key = c_key.replace('cred_', '')
             credentials[c_key] = c_value
 
+        if 'router_name' not in credentials:
+            credentials['router_name'] = provider.select_router()
+
```
lenards: I wouldn't find it intuitive (though I'm wrong, likely) that a provider would have a selected router based on a credential. 

Maybe I don't understand `ProviderCredential`?

core/models/provider.py
```diff
@@ -149,12 +150,73 @@ def is_active(self):
...
+        """
+        public_routers = self.get_credential('public_routers')
+        router_name = self.get_credential('router_name')
+        if public_routers:
+            return public_routers.split(',')
+        else:
+            return [router_name]
+
+    def select_router(self, router_distribution=None):
+        if not router_distribution:
```
lenards: Is this intended as: 
> Given a potential _weight_ distribution, pick a router out of the list of routers?

variables.ini.dist
```diff
@@ -122,6 +122,7 @@ OPENSTACK_DEFAULT_ROUTER = ; openstack_router
 OPENSTACK_USER_ROLE = ; defaultMemberRole
 OPENSTACK_ADMIN_ROLE = ; admin
 ATMOSPHERE_SUPERUSER = ; atmo_superuser
+ATMOSPHERE_SECRET_SEED = ; (Random value:) 9oe5q7cp714zmmjl51
```
lenards: Would we benefit by describing how this is created or generated? Is it just an SHA1 fingerprint? Is there any expected length to the sequence?

lenards:
@steve-gregory After reviewing the code, it seems that since we have a virtual-private-network for a given Identity - that's how/why we associated a specified router with their credentials. Am I right?

core/models/provider.py
steve-gregory: Not quite, a router distribution is the _current count of users to router_.

The 'select_router' function is designed to place the user on the router with the *smallest* count to get a 'roughly equal' distribution.

It is able to be passed in as a kwarg to avoid calling `get_router_distribution()` on each iteration of the admin script.

variables.ini.dist
steve-gregory: Random Value, A-Z and 0-9 required. Any length of string is fine.

I will add this to the comments.. I dont think we need a generator at this level.

core/models/identity.py
steve-gregory: This line is supposed to ensure that all ​*new*​ identities are created with a `router_name` at the identity level.

Currently, when `service/accounts/openstack.py` calls create_identity, it does not explicitly create a router name.

The ​*old*​ behavior was to define the router name at the ​*provider*​ level.
The ​*new*​ behavior will defer to the old behavior for 'old' identities, but any identity that has a `router_name` assigned will override the value.

scripts/admin_redistribute_routers.py
```diff
@@ -0,0 +1,44 @@
...
+    redistribute = args.redistribute
+    for provider in Provider.objects.filter(id=provider_id):
+        if redistribute:
+            needs_router = provider.identity_set.all()
+        else:
+            needs_router = provider.missing_routers()
+
+        router_map = provider.get_router_distribution(redistribute=redistribute)  # Print 'before'
+        if redistribute:
+            router_map = {key: 0 for key in router_map.keys()}
```
steve-gregory: @lenards here is the router mapping that i refer to in the comment-chain above...

scripts/admin_redistribute_routers.py
```diff
@@ -0,0 +1,44 @@
...
+        for identity in needs_router:
+            # Select next available router for the identity
+            selected_router = provider.select_router(router_map)
+            # Save happens here:
+            Identity.update_credential(identity, 'router_name', selected_router, replace=True)
```
steve-gregory: Note that we only update credentials to actually assign a new router name from an admin script.

We don't want to adversely affect users by assigning a router_name before all their instances have been suspended and their tenant network has been dismantled.

This script provides the ability to redistribute the routers as part of a maintenance period where instances are stopped or suspended.

variables.ini.dist
steve-gregory: Added defaults to `secrets.py.j2`

steve-gregory:
Some general comments related to questions in the PR:

> This might be silly - but is there any benefit to making the value an actually valid syntactical Python list?

I did consider this approach, but the truth is that if we are going to *pretend that this is a list* we should really be *storing* the actual objects in a more dynamic fashion. For example:
```
class Router(models.Model)
   name = CharField(..)
   provider = ForeignKey(provider, related_name='available_routers')  # Implicit one-to-many here, so we could call: provider.available_routers.all()
```
However there are some 'tripping points' that occur when doing this:
* Linking 'Router' to 'Identity' would be **cloud-implementation specific**
* Routers that are 'disabled' at the DB level would need to be 're-configured' dynamically (Which may not be possible if instances are still running)

Additionally, it was previously discussed that:
1. Re-distribution of routers is a rare event and routers will not often be added/removed.
2. The standard behavior, to set a single router name, will still work by default. Old clouds are not affected.
3. An identity can only have a single router
4. `Credentials` already provides a way to store a single key/value pairing and pass directly into the relevant account drivers.

So the approach I decided to go with was:
For old accounts who have not been re-distributed to a router_name in the identity:
* Infer the router_name based on the required value stored in the provider

For new accounts:
* Request a router_name from the provider (Can be a single, default name or a list of routers)

If new routers are added and users are *NOT* redistributed:
* Continue adding new users to this router until it is "eventually equal"

> After reviewing the code, it seems that since we have a virtual-private-network for a given Identity - that's how/why we associated a specified router with their credentials. Am I right?

Yes, the virtual networking setup of Neutron requires, among other things:
A subnet, a network, and a router interface (link/port) to connect your subnet/net <--> a_public_router

After the router interface is connected, presuming the router can 'talk to the outside world', your instance will be routable, etc.

So we assign a specified router with each identity to help 'divide out' the work load...

variables.ini.dist
lenards: Got it. 👍

lenards:
Okay - the last line was the most important to my understanding: 
> we assign a specified router with each identity to help 'divide out' the work load...

core/models/provider.py
```diff
@@ -149,12 +150,87 @@ def is_active(self):
...
+        router_name = self.get_credential('router_name')
+        if public_routers:
+            return public_routers.split(',')
+        else:
+            return [router_name]
+
+    def select_router(self, router_distribution=None):
+        """
+        Select and return the router_name with the smallest number of users
+
```
lenards: Perfect.

lenards:
Okay. Looks good. 🎆 

:shipit:

PR #138, opened 2016-05-03
Glance name optionally includes version name

steve-gregory:
Previous versions of glance_write_method would show only `ApplicationName` after multiple versions were created, the machines would become ambiguous:
```
ApplicationName
ApplicationName
ApplicationName
```
New approach would lead to the following:
```
ApplicationName v.1.0
ApplicationName v.2.0
ApplicationName v.final-2.5
```
Additionally, the change log will be included to help aid those who are using the glance API directly in finding the most information about a specific image.

TODO:
- [x] An administration script that will "update" all machines on a provider (registered in Atmosphre) to take the new machine name form. (Optionally, takes a subset of images)


steve-gregory:
Leaving this in WIP until I can put up an OSAD cloud and test that renaming of images will work without negatively affecting "Older systems" 👍

steve-gregory:
Originally targeted for `nautical-nighthawk`.. This PR will see light another day. ☀️

PR #139, opened 2016-05-03
API v2 POST instance error handling

steve-gregory:
Fix for [ATMO-1299](https://pods.iplantcollaborative.org/jira/browse/ATMO-1299).


steve-gregory:
Tested and verified. Self :shipit:

PR #141, opened 2016-05-10
[ATMO-991] Update html emails

cdosborn:
A bug slipped through passing the `data` keyword argument to a function that no longer accepted it. Also InstanceReport request tracker now includes RT.

All three email requests were tested with success in troposphere. In the old UI, feedback was tested.


cdosborn:
fwiw, "style-wise" i changed the functions to have keyword args on the next line

steve-gregory:
:shipit:

PR #143, opened 2016-05-12
Atmo 991 fix email

cdosborn:
This is just a fix to the template files. It's a temporary fix, before html templates are in place.
Rather than `blah<colon>` now its `blah<space><colon>`.

This can be safely pulled into prod w/o a restart. From the django docs:

> By default, the templating system will read and compile your templates every time they need to be rendered.


steve-gregory:
:zzz: :shipit:

PR #145, opened 2016-05-16
Add support for 'DEFAULT_PASSWORD_LOOKUP' to set the strategy for ope…

steve-gregory:
Add support for Salted SHA256 passwords, and the ability to change what password scheme is used on a site based on the local.py configuration.

Also updated the script to allow for 'dry runs' to test all methods of password creation quickly and for all known accounts.

Looking to include into the "Second release" of `N-N`.


lenards:
I think we want call out this as a "post-merge" step: 
- [ ] defined the value for `DEFAULT_PASSWORD_LOOKUP` for all build_env(s) in secrets repository

lenards:
👍 :shipit:

PR #146, opened 2016-05-16
Expand quota to include Compute, Storage & Network values

steve-gregory:
Highlight Reel:
* Add new Compute Quota values: Instance Count
* Add new Volume Quota values: Snapshot Count & Volume Count
* Add new Network Quota values: Floating IP Count & Port Count
* Remove Suspended Count (artificial value)
* Clean up complicated logic revolving around suspended_count and 'disappearing quota'.

Once set, Quota values will be saved to the cloud via `service.tasks.admin.set_provider_quota`.

Slated for the "Second release of N".


lenards:
We probably need/want to create a DOCS ticket for this as things like the _suspended_ count (re ec4eba6) are gone and affect the current "User Manual".

.gitignore
```diff
@@ -1,7 +1,7 @@
 B.DS_Store
 *.class
-.pyc
-.pyo
+*.pyc
+*.pyo
```
lenards: 👍

core/migrations/0052_expand_quota.py
```diff
@@ -0,0 +1,39 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
```
lenards: ❤️ `__future__`

core/migrations/0053_expand_quota_pt2.py
```diff
@@ -0,0 +1,22 @@
...
+    operations = [
+        # migrations.AlterUniqueTogether(
+        #     name='quota',
+        #     unique_together=set([('cpu', 'memory', 'storage', 'floating_ip_count', 'port_count', 'instance_count', 'snapshot_count', 'storage_count')]),
+        # ),
+        # migrations.RemoveField(
+        #     model_name='quota',
+        #     name='suspended_count',
+        # ),
+    ]
```
lenards: If I'm correct, this becomes a "no-op" with the comments out `operations`. 

Is that what we want?

service/quota.py
```diff
@@ -1,16 +1,106 @@
...
 
+
+def check_over_instance_quota(
+        username, identity_uuid,
+        esh_size=None, action=None):
+    """
+    Checks quota based on current limits (and an instance of size, if passed).
+
+    return or raise exc
+    """
```
lenards: is `esh_size` the same as "instance size"? Or is the comment talking about an _object_ instance? (or type Size)

service/quota.py
```diff
@@ -1,16 +1,106 @@
...
+    membership = IdentityMembership.objects.get(
+        identity__uuid=identity_uuid,
+        member__name=username)
+    quota = membership.quota
+    identity = membership.identity
+    driver = get_cached_driver(identity=identity)
```
lenards: Would we benefit from passing `membership` into this method? It _could_ make it easier to test in the future (since it would remove the method's dependence on the data. (I don't know the usage/call-site, so that might make it not worth the abstraction)

service/quota.py
```diff
@@ -1,16 +1,106 @@
...
+    new_port = new_floating_ip = new_instance = new_cpu = new_ram = 0
+    if esh_size:
+        new_cpu += esh_size.cpu
+        new_ram += esh_size.ram
+    if action in ['launch', 'resume', 'reboot', 'unshelve']:
```
lenards: Will the list of actions change per ☁️ provider?

service/quota.py
```diff
@@ -1,16 +1,106 @@
...
+        new_snapshot_size=0, new_volume_size=0):
+    """
+    Checks quota based on current limits (and an instance of size, if passed).
+
+    return 5-tuple: ((bool) over_quota,
+                     (str) resource_over_quota,
+                     (int) number_requested,
+                     (int) number_used,
+                     (int) number_allowed)
+    """
```
lenards: It looks like this is not returning a 5-tuple - but a bool. Is that the intent? Is the comment out-of-date with the method impl?

service/quota.py
```diff
@@ -1,16 +1,106 @@
...
+    # Will throw ValidationError if false.
+    has_storage_quota(driver, quota, new_disk)
+    has_storage_count_quota(driver, quota, new_volume)
+    has_snapshot_count_quota(driver, quota, new_snapshot)
+    return True
+
+
+def set_provider_quota(identity_uuid, limit_dict=None):
+    """
+    """
```
lenards: Might be worth hinting to the structure or possible key/value(s) of `limit_dict`. I know it's getting passed onto `_limit_user_quota` - so maybe that method provides more details.

lenards: Okay - from looking at `_get_hard_limits` the structure of `limit_dict` is just two keys: `"ram", "cpu"`.

service/quota.py
```diff
@@ -21,112 +111,87 @@ def _get_hard_limits(identity):
...
-        logger.debug("Quota exceed on suspended instances: %s"
-                     % user_quota.suspended_count)
-        return (True, 'suspended instance', 1,
-                cur_suspended, user_quota.suspended_count)
-    return (False, '', 0, 0, 0)
+
+def _set_volume_quota(user_quota, identity):
+    volume_values = {
+        'volumes': user_quota.storage_count,
+        'gigabytes': user_quota.memory,
```
lenards: Is it wrong that I associate `memory` with `"ram"`?  I assume that this is really "the size of the volume, in gigabytes". Am I right?

core/migrations/0053_expand_quota_pt2.py
steve-gregory: Oops. No. good catch.

service/quota.py
steve-gregory: esh was a 'shorthand' for what could be better described as `cloud_size` or more verbosely `cloud_representation_of_size`.

service/quota.py
steve-gregory: I think what you are realizing here is that the IdentityMembership is *not* the proper place for things like `quota` or `allocation`. This will be addressed in **The Future**

service/quota.py
steve-gregory: I'm going to drop this code, it was going to be used when we were "un-counting" suspended instance.

service/quota.py
steve-gregory: Confirmed. Comment is out of date.

service/quota.py
steve-gregory: You are right. This was a bad LoC! 🔎

core/models/quota.py
```diff
@@ -107,18 +136,25 @@ def default_dict(cls):
...
         db_table = 'quota'
         app_label = 'core'
-        unique_together = ("cpu", "memory", "storage", "storage_count",
-                           "suspended_count")
+        # unique_together = ("cpu", "memory", "storage",
+        #                    "floating_ip_count",
+        #                    "port_count",
+        #                    "instance_count",
+        #                    "snapshot_count",
+        #                    "storage_count")
```
steve-gregory: FYI - Removed the unique_together command because production DB failed with "duplicate keys". We can add this in a future migration if desired, after massaging the production database.

PR #147, opened 2016-05-17
Image upload metadata

steve-gregory:
This pull gives the "Second release of N" the ability to properly pass metadata to a new `Warlock` model of `Image` included with the latest installed version of `python-glanceclient`.

Highlight Reel:

- [x] Support for Kernel ID + Ramdisk ID
- [x] Support for "arbitrary" metadata to be included on image creation.
- [x] Requires update to Chromogenic 0.2.4 
- [x] Requires an update to settings/secrets.py + atmo-extras. (Must be able to generate a dict)


core/models/machine_request.py
```diff
@@ -371,15 +371,18 @@ def get_imaging_args(self):
...
+            "parent_image_id": self.instance.source.providermachine.identifier,
             #NOTE: THERE IS AN ASSUMPTION MADE HERE!
             # ASSUMPTION: the Creator's username == the LINUX username that was also created for them!
             #FIXME if the ASSUMPTION above changes!
             "created_by": self.instance.created_by.username,
             # Helpful for debugging
             #"parent_image_id": self.instance.source.identifier,
-            #"keep_image": True,  # Helpful for debugging
-            #"upload_image": True,  # Helpful for debugging
+            "remove_image": True if not debug else False,  # Set to False to keep Snapshot or parent_image_id in glance
```
cdosborn: These (L383,L384) can be simplified to `"remove_image": not debug`

cdosborn:
I was checking the`chromogenic` dependency. Does the version of `python-glanceclient` need to get bumped [here](https://github.com/iPlantCollaborativeOpenSource/chromogenic/blob/master/requirements.txt)?

steve-gregory:
No update of glanceclient required, thanks for the check! Going to :shipit:

PR #150, opened 2016-05-20
Update settings to allow emulation by AuthToken

steve-gregory:
When changing the templates, it seems we unfairly grouped `AuthTokenLoginBackend` as a 'globus-only' requirement..


steve-gregory:
self-:shipit:

PR #151, opened 2016-05-24
Removed size limit to prevent multiple rotates/day

nfaction:
If size is low, e.g. 100M and the file grows GB/day, it will rotate the log multiple times and could drop logs off the end faster than we wish to do so.  If this configuration is not preferred, we could try to put the size back in, but bump size up to 500M or so.

### DIFF
--- a/extras/logrotate.atmosphere
+++ b/extras/logrotate.atmosphere
@@ -10,7 +10,6 @@
 }
 /opt/dev/atmosphere/logs/libcloud.log {
     daily
-    size 100M
     rotate 10
     missingok
     compress


steve-gregory:
@nfaction I'd say we leave it off or make the entire thing `.j2` configurable so it can be configured based on the environment (Because 100 is too small for production but maybe too big for small development VMs)

lenards:
👍 on this being configurable.

cdosborn:
This commit @64e5e4a0380f4e92d86c3bce71 was merged in removing libcloud size field.

PR #152, opened 2016-05-26
Monitor Volumes

steve-gregory:
First there was monitor_instances.. Then machines.. and sizes.. This was the next logical step and will break us from our "Old cloud pulling ways" :rainbow:

### DIFF
--- a/service/tasks/monitoring.py
+++ b/service/tasks/monitoring.py
@@ -2,6 +2,7 @@ from datetime import timedelta
 
 from django.conf import settings
 from django.db.models import Q, Count
+from django.core.exceptions import ObjectDoesNotExist
 from django.utils import timezone
 
 from celery.decorators import task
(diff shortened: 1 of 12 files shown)


service/tasks/monitoring.py
```diff
@@ -520,10 +511,66 @@ def monitor_instances_for(provider_id, users=None,
...
+@task(name="monitor_volumes")
+def monitor_volumes():
+    """
+    Update volumes for each active provider.
+    """
+    for p in Provider.get_active():
+        monitor_volumes_for.apply_async(args=[p.id])
+
+@task(name="monitor_volumes_for")
+def monitor_volumes_for(provider_id, print_logs=False):
```
steve-gregory: This is the code up for review. Apparently I leaked some doc-string changes into this branch :)

steve-gregory:
Self-:shipit:

PR #155, opened 2016-05-27
[ATMO-1367] Task remove_ips fails

julianpistorius:
Added missing `provider` parameter for the `update_metadata` task.


steve-gregory:
:shipit: Thanks @julianpistorius!

PR #159, opened 2016-06-08
Enable gzip asset delivery from Nginx

lenards:
This pull requests adds a file to be linked into `/etc/nginx/gzip_settings` enabled gzip functionality and states the MIME type of assets to be transparently served if available within static _locations_. 

This is primarily a user interface server concern, but since Atmosphere & Troposphere currently compete over `/etc/nginx/sites-enabled/site.conf`, I am offering this pull request to ensure that the `site.conf` has gzip enabled regardless of the order of automation task execution. 

The uncompressed assets will still be "resolve-able" via static _locations_, but the Troposphere bundling will provide `.gz` versions which will be served if a browser send the `'Accept-Encoding: gzip'` header (even [Lynx 2.6+ supports this](http://schroepl.net/projekte/mod_gzip/browser.htm)!). Nginx will respond with `'Content-Encoding: gzip'` and the browser will uncompress them.

What does this all mean? 

It means that the 1.99 MB asset `app-[hash].js` can be compressed to 524 kB, shipped over the wire, and uncompressed by the browser. 

Modified production bundle output from Webpack:
```
           app-6d31366469fb9d34f087.js    1.99 MB       
        public-6d31366469fb9d34f087.js    1.11 MB       
     public-6d31366469fb9d34f087.js.gz     294 kB      
        app-6d31366469fb9d34f087.js.gz     524 kB      
```

See [ATMO-1371](https://pods.iplantcollaborative.org/jira/browse/ATMO-1371)


extras/nginx/Makefile.j2
```diff
@@ -9,7 +9,8 @@ KEY_PATH={{ KEY_PATH }}/{{ KEY_FILE }}
 KEY_SIZE={{ KEY_SIZE }}
 SERVER_URL={{ SERVER_URL }}
 SITES_AVAILABLE_DIR={{ SITES_AVAILABLE_DIR }}
-SITES_ENABLED_DIR= {{ SITES_ENABLED_DIR }}
+SITES_ENABLED_DIR={{ SITES_ENABLED_DIR }}
+SERVER_CFG_DIR={{ SITES_ENABLED_DIR }}/..
```
steve-gregory: I'm just wondering if this will bite us somewhere.. we could always define {{SERVER_CFG_DIR}} and then `{{SERVER_CFG_DIR}}/sites-enabled` (more reliable)..

lenards: Okay - I felt a bit dirty doing this ... but we are going to re-work all of this when it moved into Clank (that's what made me throw caution to wind :dash: 🎐)

lenards: I innovented `gzip-settings` as a file name - in most examples they include the gzip directives within `site.conf`

lenards:
Note - this makes use of nginx gzip functionality (included with nearly all builds) and the gzip-static-module (which is **not** included in the `brew install nginx` formula, but can be added by tapping `homebrew/nginx` and passing `--with` flags; more information: [homebrew-nginx](https://github.com/Homebrew/homebrew-nginx). 

Given that we have lost @prosif - I am likely to be the only one affected by this.

PR #160, opened 2016-06-09
Add warning hooks

cdosborn:
This includes some hooks that are run when the user pulls from master (more
specifically any merge). They comment when there are changes made to
requirements.txt or when migration files are added.

For example:

    [requirements-warning.rb hook]: New python requirements.
    pip install dev_requirements.txt requirements.txt
    [migrations-warning.rb hook]: New migrations found.
    ./manage.py migrate

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,14 @@ your code passes all the tests.
 ln -s contrib/pre-push.hook .git/hooks/pre-push
 ```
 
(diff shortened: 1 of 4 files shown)


lenards:
This is cool! I want to get this into _troposphere_ (with checks on package.json to suggest `npm i` too)

PR #163, opened 2016-06-20
Atmo 1369 assign project on v2 create volume

steve-gregory:
[x] - Updated the `create_volume_or_fail` method to return a core object
rather than a cloud object.
[x] - Update v1 API to call `create_esh_volume`
[x] - v2 API serializer now has a POST Serializer, and swaps serializers
before returning the value.
[x] - Project appends volume at the end, it cannot be added in the save()
method.

### DIFF
--- a/api/v2/serializers/details/volume.py
+++ b/api/v2/serializers/details/volume.py
@@ -29,10 +29,6 @@ class VolumeSerializer(serializers.HyperlinkedModelSerializer):
                                  read_only=True)
 
     projects = serializers.PrimaryKeyRelatedField(many=True, read_only=True)
-    snapshot_id = serializers.CharField(write_only=True, allow_blank=True,
-                                        required=False)
-    image_id = serializers.CharField(write_only=True, allow_blank=True,
-                                     required=False)
     uuid = serializers.CharField(source='instance_source.identifier',
                                  read_only=True)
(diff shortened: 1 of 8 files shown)


api/v2/serializers/details/volume.py
```diff
@@ -55,50 +51,15 @@ class Meta:
...
-            "instance_source": source,
-            "start_date": start_date
-        }
-
-        return Volume.objects.create(**kwargs)
+        if not data and not self.initial_data:
+            return data
+        raise Exception(
+            "This serializer for GET output ONLY! -- "
+            "Use the POST or UPDATE serializers instead!")
```
cdosborn: This block is a little confusing to me. Idk the details of `Serializers` shouldn't we just raise exception if this method is called. I.e Why would validate ever get called w/o `data` seems hacky.

api/v2/serializers/post/volume.py
```diff
@@ -0,0 +1,87 @@
+from core.models import (
+    InstanceSource, Volume, Identity, Project)
+from rest_framework import serializers
+
+
+class VolumeSerializer(serializers.ModelSerializer):
+    """
+    This serializer should only be used for Volume Creation.
+    This serializer should *never* be returned to the user.
+    Instances created from this serializer should be
```
cdosborn: Should this be "Volumes created..."?

PR #165, opened 2016-06-24
Allow user to 'complete the loop' and close their own Requests

steve-gregory:
Authors of Machine/Resource Requests should now be able to close their own requests within the API.

 (NOTE: Intentionally 'closes' the request, rather than delete.)

REF: [JIRA](https://pods.iplantcollaborative.org/jira/browse/ATMO-1408)

### DIFF
--- a/api/v2/views/base.py
+++ b/api/v2/views/base.py
@@ -24,9 +24,16 @@ def unresolved_requests_only(fn):
     @wraps(fn)
     def wrapper(self, request, *args, **kwargs):
         instance = self.get_object()
-        user_can_act = request.user.is_staff or request.user.is_superuser or CloudAdministrator.objects.filter(user=request.user.id).exists()
-        #TODO: Logic needs 're-worked' here. MachineRequests in 'non-final' states should be allowed to be PATCH'ed for re-submission.
-        if (not user_can_act and hasattr(instance, "is_closed") and instance.is_closed()):
+        staff_can_act = request.user.is_staff or request.user.is_superuser or CloudAdministrator.objects.filter(user=request.user.id).exists()
+        user_can_act = request.user == getattr(instance, 'created_by')
+        if not user_can_act and not staff_can_act:
+            message = (
+                "Method '%s' not allowed: "
+                "Only staff members and the owner are authorized to make this request."
+                % self.request.method
+            )
+            raise exceptions.NotAuthenticated(detail=message)
+        if (not staff_can_act and hasattr(instance, "is_closed") and instance.is_closed()):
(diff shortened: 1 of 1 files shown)


api/v2/views/base.py
```diff
@@ -24,9 +24,16 @@ def unresolved_requests_only(fn):
...
+        staff_can_act = request.user.is_staff or request.user.is_superuser or CloudAdministrator.objects.filter(user=request.user.id).exists()
+        user_can_act = request.user == getattr(instance, 'created_by')
+        if not user_can_act and not staff_can_act:
+            message = (
+                "Method '%s' not allowed: "
+                "Only staff members and the owner are authorized to make this request."
+                % self.request.method
+            )
+            raise exceptions.NotAuthenticated(detail=message)
+        if (not staff_can_act and hasattr(instance, "is_closed") and instance.is_closed()):
```
julianpistorius: Possibly modify to just test for `instance.is_closed()`.

Also write a comment to explain this logic in English.

api/v2/views/base.py
```diff
@@ -157,8 +164,24 @@ def perform_create(self, serializer):
...
+            }
+            logger.exception(e)
+            raise exceptions.ParseError(detail=message)
+
+    def perform_destroy(self, instance):
+        """
+        Add an end date to a request and take no further action
+        """
+        status, _ = StatusType.objects.get_or_create(name="closed")
+        instance.status = status
```
julianpistorius: Note to self: StatusType is specifically for BaseRequest.

julianpistorius:
Looks good! Thanks @steve-gregory.

cdosborn:
I don't agree with this pr.

`perform_destroy` should delete the model, this is what someone would expect coming from django. There is already the mechanism to `close` [here](https://github.com/steve-gregory/atmosphere/blob/fffe5cb32203491a699434efc382f3f29957fe28/api/v2/views/base.py#L210)

Lines 177-182 should be removed. The issue should be resolved in troposphere to patch a `BaseRequest` with status closed.

api/v2/views/base.py
cdosborn: This is okay since, `Machine/Resource` are subtypes, and so far we havn't needed statuses that are unique to each.

cdosborn:
After chatting with @steve-gregory, it is more consistent with our service to `end_date` rather than delete in the `perform_destroy`. I think this is good perhaps a little dissonant, but essentially 'closed' is synonymous with `destroyed`. 👍

steve-gregory:
:shipit:

PR #166, opened 2016-06-29
Fix templates for support to allow quotes, brackets, etc.

cdosborn:
TLDR: let the email ticket system be responsible for sanitizing support emails

By default django auto-escapes all templates. You have to mark something as
safe for it to be not escaped. Our issue is that django was escaping html,
and then RT escaped the escaped content!

**note:** these templates were tested!

### DIFF
--- a/core/templates/core/email/feedback.html
+++ b/core/templates/core/email/feedback.html
@@ -1,6 +1,6 @@
-Username : {{ username }} ({{ name }})
+Username : {{ username }} ({{ name|safe }})
 
-Feedback : {{ feedback }}
+Feedback : {{ feedback|safe }}
 
 Provider : {{ provider }}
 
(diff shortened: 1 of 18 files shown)


core/templates/core/email/feedback.html
```diff
@@ -10,16 +10,16 @@
 
 Instances :
 {% for i in instances %}
-    Name : {{ i.name }}
-    ID : {{ i.provider_alias }}
+    Name : {{ i.name|safe }}
+    UUID : {{ i.provider_alias }}
```
cdosborn: Small name change. A model id mean something different that a models uuid

core/templates/core/email/volume_report.html
```diff
@@ -1,14 +1,14 @@
...
 {% endif %}
 Details : 
-    {{ message }}
+    {{ message|safe }}
 
 Volume :
-    Name : {{ volume.name }}
-    ID : {{ volume.id }}
+    Name : {{ volume.name|safe }}
+    UUID : {{ volume.identifier }}
```
cdosborn: We care about the volume uuid more than the model id

steve-gregory:
@cdosborn should this be pointing to `petrified-penguin`?

cdosborn:
@steve-gregory Somehow extra commits got tacked on here. I'll resubmit a pr against PP.

PR #169, opened 2016-07-10
Updated Django template backend definitions

lenards:
When upgrading from Django 1.8 to 1.9, there were [changes to the definition of settings.TEMPLATES](https://docs.djangoproject.com/en/1.9/ref/templates/upgrading/). The presence of `TEMPLATE_DEBUG` will cause a warning to generated (this will happen in atmosphere upon `service atmosphere start`). A similar PR was submitted to troposphere (as [Troposphere PR 403](https://github.com/iPlantCollaborativeOpenSource/troposphere/pull/403)).

The warning message can be produced by doing a `check`:
```
09:59 $ ./manage.py check 
System check identified some issues:

WARNINGS:
?: (1_8.W001) The standalone TEMPLATE_* settings were deprecated in Django 1.8 and the TEMPLATES dictionary takes precedence. You must put the values of the following settings into your default TEMPLATES dict: TEMPLATE_DEBUG.
```

The definition in atmosphere was altered to account for multiple backends present in `settings.TEMPLATES`. 

An example of the change being set in the `'OPTIONS'` of the Django backend, but not in the Jinja2 backend:
```
09:19 $ ./manage.py shell
....
In [1]: from atmosphere.settings import TEMPLATES

In [2]: TEMPLATES
Out[2]: 
[{'BACKEND': 'django.template.backends.jinja2.Jinja2',
  'DIRS': ['/Users/lenards/devel/atmosphere/core/templates/core/email']},
 {'APP_DIRS': True,
  'BACKEND': 'django.template.backends.django.DjangoTemplates',
  'DIRS': ['/Users/lenards/devel/atmosphere/templates'],
  'OPTIONS': {'context_processors': ['django.contrib.auth.context_processors.auth',
    'django.template.context_processors.debug',
    'django.template.context_processors.i18n',
    'django.template.context_processors.media',
    'django.template.context_processors.static',
    'django.template.context_processors.tz',
    'django.contrib.messages.context_processors.messages'],
   'debug': True}}]
```

### DJANGO_TEMPLATE_DEBUG = False, atmosphere/settings/local.py
```python
SECRETS_MODULE = secrets

globals().update(vars(sys.modules['atmosphere.settings']))

# Debug Mode
DEBUG = True




ENFORCING = False
```

### DIFF
--- a/atmosphere/settings/__init__.py
+++ b/atmosphere/settings/__init__.py
@@ -19,7 +19,6 @@ from kombu import Exchange, Queue
 
 # Debug Mode
 DEBUG = True
-TEMPLATE_DEBUG = DEBUG
 
 # Enforcing mode -- True, when in production (Debug=False)
 ENFORCING = not DEBUG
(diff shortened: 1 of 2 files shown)


steve-gregory:
:shipit:

PR #172, opened 2016-07-15
Fix templates for support to allow quotes, brackets, etc.

cdosborn:
TLDR: let the email ticket system be responsible for sanitizing support emails

By default django auto-escapes all templates. You have to mark something as
safe for it to be not escaped. Our issue is that django was escaping html,
and then RT escaped the escaped content!

### DIFF
--- a/core/templates/core/email/volume_report.html
+++ b/core/templates/core/email/volume_report.html
@@ -1,14 +1,14 @@
-Username : {{ username }} ({{ name }})
+Username : {{ username }} ({{ name|safe }})
 {% if problems %}
 Problems : 
-    {{ problems|join("\n")|indent(4, false) }}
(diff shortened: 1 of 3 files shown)


cdosborn:
@steve-gregory merge this when you get the chance.

core/templates/core/email/volume_report.html
```diff
@@ -1,14 +1,14 @@
...
-    {{ problems|join("\n")|indent(4, false) }}
+    {{ problems|join("\n")|indent(4, false)|safe }}
 {% endif %}
 Details : 
-    {{ message }}
+    {{ message|safe }}
 
 Volume :
-    Name : {{ volume.name }}
-    ID : {{ volume.id }}
```
steve-gregory: 👍

steve-gregory:
:shipit:

PR #173, opened 2016-07-17
Advanced provider configuration

steve-gregory:
Created a new JSONField for Provider (Requires an upgrade to Postgresql >=9.4 ++ Django 1.9 on the server dependencies) called `cloud_config`.

This field will allow us the versatility we need to allow each provider to have all of the flexibility required to avoid the traditional use of `settings.XYZ`.

Providers without a cloud_config will still fall back to the original settings, and you will see a `logger.warn` call at each point, informing the installer that the required key for cloud config was missing.

Additionally, the ./scripts/add_new_provider has been overhauled to ask even more questions to ensure your provider is properly configured.

`service/networking.py` will see new life as the holder of the two NetworkTopology/Strategy selections that are being used by cloud operators of atmosphere. Currently we support ExternalRouter configuration and ExternalNetwork configuration.

*Work yet to be completed:*
- [x] A valid cloud config example file will be completed prior to the PR completion.
- [x] The scripts/add_new_provider network selection should ask for the name of the strategy to use (and show the available list..)


api/v2/serializers/details/resource_request.py
```diff
@@ -194,6 +194,13 @@ class Meta:
         )
 
 
+def _get_pending_status():
+    try:
+        return StatusType.objects.get(name="pending")
+    except Exception:
```
cdosborn: Lets do an 'except DoesNotExist as e' so we don't catchall.
1. As a good habit
2. MultipleObjectsReturned would notify us that our db is messed up

scripts/add_new_provider.py
```diff
@@ -72,6 +81,12 @@ def review_information(provider_info, admin_info, provider_credentials):
             print "3. Provider Credentials deleted"
 
 
+def yes_no_truth(raw_text):
+    if raw_text.lower().strip() == 'yes':
```
cdosborn: This can just be `return raw_text.lower().strip() == 'yes'`. I don't think this method should exist.

scripts/add_new_provider.py
```diff
@@ -10,33 +10,40 @@
...
 KVM = PlatformType.objects.get_or_create(name='KVM')[0]
 XEN = PlatformType.objects.get_or_create(name='Xen')[0]
 openstack = ProviderType.objects.get_or_create(name='OpenStack')[0]
 
-valid_url = URLValidator()
+url_validator = URLValidator()
 
 
-def require_input(question, validate_answer=None):
+def require_input(
```
cdosborn: I think we could get by with just a small tweak to the original interface:

`require_input(input, default=None, validator=None)`. I cannot think of any edge cases that it wouldn't cover. Looking at the new interface it allows for conflicting behavior like providing a validator but setting use_validate_answer to false. I'd be willing to contribute a patch for this. Comments?

scripts/add_new_provider.py
```diff
@@ -181,12 +205,9 @@ def get_provider_info(provider_info={}):
         # 3.  Collect provider type
         print "Select a provider type for your new provider"
         print "1: Openstack"
-        while True:
-            provider_type = raw_input("Select a provider type [1]: ")
-            #NOTE: this will be replaced with actual logic when necessary.
-            if True:
-                provider_type = openstack
-                break
+        provider_type = require_input("Select a provider type [1]", default='1')
```
cdosborn: Should this be `provider_type = require_input("Select a provider type [1]", default=openstack)`?

scripts/add_new_provider.py
```diff
@@ -206,32 +227,159 @@ def get_admin_info(admin_info={}):
...
+    #get/set domain
+    domain = user_config.get('domain')
+    if not domain:
+        print "What is the domain name for your provider? (Default: default)"
+        domain = require_input("domain name for the provider: ", default='default')
+
+    #get/set password_lookup
+    password_lookup = user_config.get('password_lookup')
+    password_salt = user_config.get('password_salt')
+    # TODO: More information on what password_lookup does..
```
cdosborn: This line can be removed.

service/__init__.py
```diff
@@ -1,8 +1,15 @@
...
-import service.tasks.monitoring
-import service.tasks.driver
-import service.tasks.volume
-import service.tasks.machine
-import service.tasks.snapshot
-import service.tasks.admin
-import chromogenic.tasks
+import django
+try:
+    #FIXME: Find a way to make tasks 'visible' without abusing the __init__ files
```
cdosborn: Is this FIXME still relevant? Is it the last `chromogenic.tasks` thats triggering this?

service/accounts/openstack_manager.py
```diff
@@ -292,12 +229,20 @@ def init_keypair(self, username, password, project_name):
         return self.get_or_create_keypair(username, password, project_name,
                                           keyname, public_key)
 
-    def init_security_group(self, username, password, project_name,
-                            security_group_name, rules_list):
```
cdosborn: NIce!

service/accounts/openstack_manager.py
```diff
@@ -413,7 +359,11 @@ def accept_shared_image(self, glance_image, project_name):
     def rebuild_security_groups(self, core_identity, rules_list=None):
         creds = self.parse_identity(core_identity)
         if not rules_list:
-            rules_list = self.MASTER_RULES_LIST
+            try:
+                rules_list = self.cloud_config['network']['default_security_rules']
+            except KeyError:
+                logger.warn("Cloud config ['user']['default_security_rules'] is missing -- using deprecated settings.DEFAULT_RULES")
```
cdosborn: I think this should be:
```diff
- logger.warn("Cloud config ['user']['default_security_rules'] is missing
+ logger.warn("Cloud config ['network']['default_security_rules'] is missing
```

service/accounts/openstack_manager.py
```diff
@@ -485,33 +435,124 @@ def delete_security_group(self, identity):
...
+        else:
+            NetworkTopologyStrategyCls = get_topology_cls(topology_name)
+
+        try:
+            network_strategy = NetworkTopologyStrategyCls(identity)
+        except:
+            raise
+            raise Exception(
+                "Error initializing Network Topology - %s + %s " %
+                (NetworkTopologyStrategyCls, identity))
```
cdosborn: The second raise will never be raised. Convert the second exception to a print statement then simple `raise` after?

service/accounts/openstack_manager.py
```diff
@@ -485,33 +435,124 @@ def delete_security_group(self, identity):
...
-            username, project_name, remove_network=remove_network)
+        neutron = self.get_openstack_client(identity, 'neutron')
+        try:
+            topology_name = self.cloud_config['network']['topology']
+        except Exception:
+            logger.exception(
+                "Network topology not selected -- "
+                "Will attempt to use the last known default: ExternalRouter.")
+            topology_name = None
+        network_strategy = self.select_network_strategy(identity, topology_name)
```
cdosborn: What happens here if `topology_name = None`?  Should `topology_name` be given a default?

service/accounts/openstack_manager.py
```diff
@@ -536,45 +577,56 @@ def os_delete_account(self, username, projectname):
...
+            try:
+                strategy = self.cloud_config['user']['password_lookup']
+            except KeyError:
+                logger.warn("Cloud config ['user']['password_lookup'] is missing -- using deprecated settings.DEFAULT_PASSWORD_LOOKUP")
+                strategy = DEFAULT_PASSWORD_LOOKUP
 
         if not strategy\
                 or strategy == 'old_hashpass':
-            return cls.old_hashpass(username)
+            return self.old_hashpass(username, secret_salt)
```
cdosborn: `old_hashpass` does not accept a salt as an argument.

service/accounts/openstack_manager.py
```diff
@@ -536,45 +577,56 @@ def os_delete_account(self, username, projectname):
...
         """
         Create a unique password using 'Username' as the wored
         and the SECRET_KEY as your salt
         """
+        # Get salt from config or settings
+        try:
+            secret_salt = self.cloud_config['user']['password_salt']
+        except KeyError:
+            logger.warn("Cloud config ['user']['password_salt'] is missing -- using deprecated secrets.SECRET_SEED")
+            secret_salt = SECRET_SEED
```
cdosborn: What if `SECRET_SEED` is None. Why are we hashing usernames \o . o/

service/accounts/openstack_manager.py
```diff
@@ -536,45 +577,56 @@ def os_delete_account(self, username, projectname):
...
         return sha1(username).hexdigest()
 
-    @classmethod
-    def salt_hashpass(cls, username):
+    def salt_hashpass(self, username, secret_salt):
         from hashlib import sha256
-        secret_salt = SECRET_SEED.translate(None, string.punctuation)
+        if not secret_salt:
+            secret_salt = SECRET_SEED.translate(None, string.punctuation)
         password = sha256(secret_salt + username).hexdigest()
```
cdosborn: Unless i'm misunderstanding this:
```diff
- password = sha256(secret_salt + username).hexdigest()
+ password = sha256(secret + salt + username).hexdigest()
```

If we're using a `cloud_config` password per provider, the regular `hash_pass` would be:
```
password = sha256(cloud_pass + username).hexdigest()
```

If we want to salt it with our application secret (which we should just always do:
```
password = sha256(cloud_pass + SECRET_SEED + username).hexdigest()
```

scripts/add_new_provider.py
steve-gregory: We could, but we could also change what this context means in the future (I.e. `in ['y','yes',...]`)

scripts/add_new_provider.py
steve-gregory: Sure

service/accounts/openstack_manager.py
steve-gregory: Yes!

service/accounts/openstack_manager.py
steve-gregory: Yes, that would be appropriate

service/accounts/openstack_manager.py
steve-gregory: This mechanism is no longer used. It should be removed.

service/accounts/openstack_manager.py
```diff
@@ -485,33 +435,124 @@ def delete_security_group(self, identity):
...
 
-    def delete_network(self, identity, remove_network=True):
-        # Core credentials need to be converted to openstack names
+    def select_network_strategy(self, identity, topology_name=None):
+        """
+        Select a network topology and initialize it with the identity/provider specific information required.
+        """
+        # Select Cls
+        if not topology_name:
+            NetworkTopologyStrategyCls = ExternalRouter
```
steve-gregory: @cdosborn this is the default.

service/accounts/openstack_manager.py
steve-gregory: First raise is a leftover from debugging, it should be removed.

api/v2/serializers/details/resource_request.py
steve-gregory: You've been invited as a collaborator

scripts/add_new_provider.py
steve-gregory: Thanks @cdosborn I think this could be retooled to use less ages and get the same desired behavior

service/instance.py
```diff
@@ -268,11 +268,15 @@ def detach_port(esh_driver, esh_instance):
     return result
 
 
-def remove_network(esh_driver, identity_uuid, remove_network=False):
+def remove_network(esh_driver, identity_uuid):
```
cdosborn: Remove `remove_empty_network` only does something if there are no active instances. By association `remove_network` only does something if there are no active instances (doesn't sit well with my intuition). 

provider actions like `stop_instance` Do the following:
```
    if reclaim_ip:
        remove_ips(esh_driver, esh_instance)
    stopped = esh_driver.stop_instance(esh_instance)
    if reclaim_ip:
        remove_network(esh_driver, identity_uuid)
```

It would be better like so:
```diff
    if reclaim_ip:
        remove_ips(esh_driver, esh_instance)
    stopped = esh_driver.stop_instance(esh_instance)
    if reclaim_ip:
-       remove_network(esh_driver, identity_uuid)
+      # Cleanup network if this is the last network
+      remove_empty_network(esh_driver, identity_uuid)
```

To be clear, i'm proposing renaming this to: `remove_empty_network`

service/networking.py
```diff
@@ -1,8 +1,303 @@
...
+    the related math.
+    """
+    MAX_SUBNET = 4064  # Note 16 * 256
+    n = uid_number % MAX_SUBNET
+
+    #16-31
+    block1 = (n + inc) % 16 + 16
+
+    #1-254
+    block2 = ((n + inc) / 16) % 254 + 1
```
cdosborn: For my own sake:
`% 254` forces [0, 253]
`+1` forces [1, 254]

Why are you dividing by 16? It should still achieve what you want?

service/accounts/openstack_manager.py
```diff
@@ -28,76 +21,21 @@
...
-from core.ldap import get_uid_number
 from core.models.identity import Identity
 
 from service.accounts.base import BaseAccountDriver
+from service.networking import get_topology_cls, ExternalRouter, ExternalNetwork
 
 from atmosphere.settings.secrets import SECRET_SEED
-from atmosphere.settings import DEFAULT_PASSWORD_LOOKUP, DEFAULT_PASSWORD_UPDATE
-
-def get_unique_id(userid):
```
cdosborn: This is deleted def is referenced [here](https://github.com/steve-gregory/atmosphere/blob/1d6053d1e7db08cf5002e6331ad2465518aaccd7/service/accounts/openstack_manager.py#L422)

service/networking.py
```diff
@@ -1,8 +1,303 @@
...
+    Return the default subnet for the username and provider.
+
+    Add and mod by inc to allow for collitions.
+    """
+    if get_uid_number:
+        uid_number = get_uid_number(username)
+    else:
+        uid_number = 0
+
+    if uid_number:
```
cdosborn: This `if` block can be replaced by `(block1, block2) = get_ranges(uid_number, inc)`

service/networking.py
```diff
@@ -1,8 +1,303 @@
...
+                                "for user: %s" % username)
+                    inc += 1
+            except NeutronClientException as nce:
+                if "overlap" in nce.message:
+                    # expected output. hash already use, add one and try another subnet.
+                    inc += 1
+                else:
+                    logger.exception("Unable to create subnet for user: %s" % username)
+                    inc += 1
+                if not get_unique_number:
```
cdosborn: This check should be performed above.

service/networking.py
```diff
@@ -1,8 +1,303 @@
...
+                    logger.warn("No get_unique_number method "
+                                "provided for user: %s" % username)
+            except Exception as e:
+                logger.exception("Unable to create subnet for user: %s" % username)
+                if not get_unique_number:
```
cdosborn: This can be deleted with this [fix](https://github.com/iPlantCollaborativeOpenSource/atmosphere/pull/173/files#r71093510)

service/networking.py
```diff
@@ -1,8 +1,303 @@
...
+                if cidr:
+                    return network_driver.create_subnet(neutron, subnet_name,
+                                              network_id, ip_version,
+                                              cidr, dns_nameservers)
+                else:
+                    logger.warn("Unable to create cidr for subnet "
+                                "for user: %s" % username)
+                    inc += 1
+            except NeutronClientException as nce:
+                if "overlap" in nce.message:
```
cdosborn: I did a quick git grep of their code base. I could not find "overlap" in any of their raised exceptions. They raise specific exceptions, if there's a specific one, we should rely on that rather than parsing the message, which isn't an interface we can depend on.

service/networking.py
```diff
@@ -1,8 +1,303 @@
-from atmosphere.settings import ATMOSPHERE_PRIVATE_KEYFILE
+"""
+Atmosphere network
+
+This file will help a driver infer the strategy to use when creating or destroying a "project network".
+"""
+import random
+
+from neutronclient.common.exceptions import NeutronClientException, NotFound
```
cdosborn: `neutronclient` isn't in our `requirements.txt` it looks like it was remove in 2014?

service/networking.py
```diff
@@ -1,8 +1,303 @@
...
+
+    def validate(self, core_identity):
+        identity_creds = core_identity.get_all_credentials()
+        if 'network_name' not in identity_creds.keys():
+            logger.warn("Credential 'network_name' missing:"
+                        "cannot create user network")
+            raise Exception("Identity %s has not been assigned a 'network_name'" % core_identity)
+        return True
+
+    def delete_network(self, network_driver, user_neutron, network_name):
```
cdosborn: Should this method be removed to defer to the parent class?

service/networking.py
```diff
@@ -1,8 +1,303 @@
...
+        return None
+
+    def get_or_create_network(self, network_driver, user_neutron, network_name):
```
cdosborn: Should this create a network if it doesn't exist?

service/accounts/openstack_manager.py
cdosborn: I see. This could be changed to:
```
        topology_name = getattr(self.cloud_config['network'], 'topology', None)

        network_strategy = None
        if not topology_name:
            logger.exception(
                "Network topology not selected -- "
                "Will attempt to use the last known default: ExternalRouter.")
            network_strategy = self.select_network_strategy(identity)
        else:
            network_strategy = self.select_network_strategy(identity, topology=topology_name)
```

It's easier to verify visually that `topology_name=None` will not have a bad impact on `select_network_strategy`.

service/accounts/openstack_manager.py
cdosborn: 👓

service/accounts/openstack_manager.py
```diff
@@ -485,33 +435,124 @@ def delete_security_group(self, identity):
...
         identity_creds = self.parse_identity(identity)
-        username = identity_creds["username"]
         project_name = identity_creds["tenant_name"]
-        # Convert from libcloud names to openstack client names
-        return self.network_manager.delete_project_network(
-            username, project_name, remove_network=remove_network)
+        neutron = self.get_openstack_client(identity, 'neutron')
+        try:
+            topology_name = self.cloud_config['network']['topology']
+        except Exception:
```
julianpistorius: KeyError?

scripts/add_new_provider.py
```diff
@@ -206,32 +227,159 @@ def get_admin_info(admin_info={}):
...
+        "Do users need to create their own network Yes/[No]",
+        yes_no_truth, default='no', allow_falsy=True, use_validated_answer=True)
+        user_subnet = require_input(
+        "Do users need to create their own subnet Yes/[No]",
+        yes_no_truth, default='no', allow_falsy=True, use_validated_answer=True)
+    else:
+        topology = net_config['topology']
+        user_router = topology.get('user_router',False)
+        user_network = topology('user_network',False)
+        user_subnet = topology('user_subnet',False)
```
julianpistorius: Should these be:

```diff
-        user_network = topology('user_network',False)
-        user_subnet = topology('user_subnet',False)
+        user_network = topology.get('user_network',False)
+        user_subnet = topology.get('user_subnet',False)
```

service/accounts/openstack_manager.py
julianpistorius: Not sure about `getattr` there @cdosborn. Isn't `cloud_config['network']` a dictionary? Don't you mean:

```python
topology_name = self.cloud_config['network'].get('topology')
```

service/accounts/openstack_manager.py
```diff
@@ -485,33 +435,124 @@ def delete_security_group(self, identity):
...
-            password,
-            project_name,
-            get_unique_number=get_unique_id,
-            dns_nameservers=dns_nameservers,
-            **net_args)
+        neutron = self.get_openstack_client(identity, 'neutron')
+        dns_nameservers = self.dns_nameservers_for(identity)
+        try:
+            topology_name = self.cloud_config['network']['topology']
+        except Exception:
```
julianpistorius: `KeyError` for clarity?

service/accounts/openstack_manager.py
cdosborn: ha, do i even python.

service/accounts/openstack_manager.py
cdosborn: Resolved here: @4d854ca895

scripts/add_new_provider.py
steve-gregory: This could also be done as a separate PR

service/accounts/openstack_manager.py
steve-gregory: Lies! Its `_get_unique_id` its *totally different* 😀

service/networking.py
steve-gregory: This whole thing needs re-written. its a port from https://github.com/iPlantCollaborativeOpenSource/rtwo/blob/0.3/rtwo/drivers/openstack_network.py#L456-L497

service/networking.py
steve-gregory: This whole thing needs re-written. its a port from https://github.com/iPlantCollaborativeOpenSource/rtwo/blob/0.3/rtwo/drivers/openstack_network.py#L456-L497

service/networking.py
steve-gregory: No. This is the selection made by the strategy (No-op, pass, do nothing)

service/networking.py
steve-gregory: No. In an "External Network" topology failing to have the external network name results in an incomplete/invalid topology. Atmosphere cannot "Externalize a network".

cdosborn: 👓

lenards:
Since this will move us to PostgreSQL 9.4+ - we will need to address that in Clank as well. Since 9.4 is not available within the apt repo, we'll need to add to the sources list. 

The process is outlined here: https://www.postgresql.org/download/linux/ubuntu/

cdosborn:
@steve-gregory settings files need to be updated to match all the new variables that the `rtwo` bump introduced.

cdosborn:
I submitted a [PR](https://github.com/steve-gregory/atmosphere/pull/1) against @steve-gregory fork, to isolate the changes from this conversation. It fixes some of the concerns I mentioned about the `hashpass` method.

docs/NETWORKING.md
```diff
@@ -0,0 +1,29 @@
...
+
+### External Network Topology
+In an external network topology, the network looks something like this:
+```
+    INSTANCE --> User-Subnet --> External Network --> User Interface between network-router --> User-Router --> Router-gateway --> Internet
+```
+
+In order to use an external network, your provider must have a kwarg: `external_network_name`
+
+All user subnets will be attached to the external network. Additionally, one interfaces will be created between the external network and the user router, and another interface (Router Gateway) will connect the users router to the outside world.
```
cdosborn: 👍 👍 👍 👍 👍

cdosborn:
It may also be noted that LEADERBOARD got the 🔪  today.
💀 

This PR looks good to me. Thank you for fixing the imports on libcloud.

PR #177, opened 2016-07-22
Save images with application and version name

steve-gregory:
Previous versions of glance_write_method would show only ApplicationName after multiple versions were created, the machines would become ambiguous:

ApplicationName
ApplicationName
ApplicationName
New approach would lead to the following:

ApplicationName v.1.0
ApplicationName v.2.0
ApplicationName v.final-2.5
Additionally, the change log will be included to help aid those who are using the glance API directly in finding the most information about a specific image.

TODO:
- [x] An administration script that will "update" all machines on a provider (registered in Atmosphre) to take the new machine name form. (Optionally, takes a subset of images)
- [x] Verify that the administrative script works as expected on cloudlab or another isolated cloud _prior_ to merge.


api/v2/views/base.py
```diff
@@ -24,9 +24,16 @@ def unresolved_requests_only(fn):
     @wraps(fn)
     def wrapper(self, request, *args, **kwargs):
```
erbriones: Looks like this wrapper function I initially created is getting complicated.  It really was only meant to check for requests that were in a non final state. Ideally, any permission level concerns should be handler elsewhere. Just thought I would share my thoughts.

steve-gregory:
This PR still pending review until we can test this on a fresh cloud.

steve-gregory:
Before:
```
(atmo_master) root@atmosphere:/opt/dev/atmosphere# glance image-list
+--------------------------------------+---------------------------------------+
| ID                                   | Name                                  |
+--------------------------------------+---------------------------------------+
| 5fa21257-200e-4fef-85b7-0bd62c8e9450 | centos-6                              |
| 904a44e6-3fd7-4174-9acf-57b472e2346c | centos-7                              |
| 1d7bf56b-2d13-4b8c-923a-928cd388d6e7 | cyverse-centos-6.8-xfce-gui           |
| d06ff318-51af-4b8e-8953-d17857fed992 | novnc-gateway-after-host-provisioning |
| 20d1ebc9-7cb6-4a54-932f-132bbb997779 | Steves Best Image Yet                 |
| ca77727d-d30a-45c8-88f0-010f04ee58d3 | Steves Best Image Yet                 |
| f4c63e87-3479-4d26-b8b1-55abbdfaca9d | Totally Different Image               |
| 54fcdb6d-b9a8-410b-8a62-2d57cc9ff516 | ubuntu-14.04                          |
| d736d76a-253d-448f-ba3f-405d742eeb42 | ubuntu-16.04                          |
+--------------------------------------+---------------------------------------+
```
After:
```
(atmo_master) root@atmosphere:/opt/dev/atmosphere# glance image-list
+--------------------------------------+---------------------------------------+
| ID                                   | Name                                  |
+--------------------------------------+---------------------------------------+
| 5fa21257-200e-4fef-85b7-0bd62c8e9450 | centos-6 v.1.0                        |
| 904a44e6-3fd7-4174-9acf-57b472e2346c | centos-7 v.1.0                        |
| 1d7bf56b-2d13-4b8c-923a-928cd388d6e7 | cyverse-centos-6.8-xfce-gui v.1.0     |
| d06ff318-51af-4b8e-8953-d17857fed992 | novnc-gateway-after-host-provisioning |
| 20d1ebc9-7cb6-4a54-932f-132bbb997779 | Steves Best Image Yet v.1.0           |
| ca77727d-d30a-45c8-88f0-010f04ee58d3 | Steves Best Image Yet v.2.0           |
| f4c63e87-3479-4d26-b8b1-55abbdfaca9d | Totally Different Image v.3.0         |
| 54fcdb6d-b9a8-410b-8a62-2d57cc9ff516 | ubuntu-14.04 v.1.0                    |
| d736d76a-253d-448f-ba3f-405d742eeb42 | ubuntu-16.04 v.1.0                    |
+--------------------------------------+---------------------------------------+
```

cdosborn:
Names are not unique and they are also unstructured. 

If I name my application "My favorite version is v.3.0" , then that last word will be interpreted to be something it's not. I don't think glance-list output warrants this change.

steve-gregory:
@cdosborn We have to consider that the images created on atmosphere *will necessarily be available to the API user* -- It has already been brought to our attention that different versions of images that atmosphere manages seem to be "identical" and that adding version information to the name would be seen as helpful/more attractive to those API end-users.

I agree that we need an elegant way to handle version separation inside the name though. We could potentially switch separators to something less likely to cause conflict, or avoid making changes to images that are _already_ in the glance image-list and focus on images created in the future

cdosborn:
Does this mean that all images will get renamed? What about images that already have a version in the name. Maybe I'm not fully understanding the scope here.

service/openstack.py
```diff
@@ -85,13 +88,16 @@ def glance_update_machine_metadata(provider_machine, metadata={}):
...
-        "application_version": str(provider_machine.application_version.name),
+        "application_version": str(version.name),
         "application_uuid": base_app.uuid,
         "application_name": _make_safe(base_app.name),
         "application_owner": base_app.created_by.username,
         "application_tags": json.dumps(
             [_make_safe(tag.name) for tag in base_app.tags.all()]),
-        "application_description": _make_safe(base_app.description)}
+        "application_description": _make_safe(base_app.description),
+        "version_name": str(version.name),
```
steve-gregory: agreed, we should move away from `application_version` (jargon-y) in favor of `version_name`.

core/models/machine.py
```diff
@@ -319,9 +353,7 @@ def get_or_create_provider_machine(image_id, machine_name,
     if not version:
         version = get_version_for_machine(provider_uuid, image_id, fuzzy=True)
     if not version:
-        version = create_app_version(app, "1.0", provider_machine_id=image_id)
-    #TODO: fuzzy=True returns a list, but call comes through as a .get()?
-    #      this line will cover that edge-case.
+        version = create_app_version(app, version_name, provider_machine_id=image_id)
```
steve-gregory: This works as expected:
```
if version passed in, don't attempt to retrieve the version
if no version was passed in *and* attempt to retrieve version was empty, create a version.
```
To make this clearer, i will tab-over the second if statement.

steve-gregory:
Additional context (hopefully) for @cdosborn : The goal of the PR is to make it easier for _api users_ who look at a catalog that has been _actively managed via atmosphere_ to find the specific image/machine they are looking for when querying the APIs.

It's true that they **could** individually `glance image-show <...>` for each image that was listed `ubuntu 16.04`, but by including version names we can _hopefully_ avoid ambiguity in our managed catalog.

This will _not_ affect how the GUI or atmosphere API render the application, version, or provider machine. It is purely aesthetic for the "Native cloud API" user.

cdosborn:
Okay, I didn't drill down into each field to see how they were being used. Is `application_name` where we store troposphere's idea of the name? Thanks for the context.

steve-gregory:
@cdosborn yes, that is correct.

service/openstack.py
cdosborn: Not referring to the jargon, as so much the duplicate state.

PR #180, opened 2016-08-03
Add a new task in the deployment chain -- check web_desktop

steve-gregory:
WIP

Migrations are included but not added -- Will wait on this PR until AllocationSource migrations are created, to avoid re-numbering migration conflicts.


steve-gregory:
Nothing to see here until the WIP tag is removed.. Just bringing people in to the ➿

lenards:
What's left for this to get merged? I think it needs to be merged after the allocation-source work, right?

lenards:
I assume that is going to add metadata on the instance, like `{"novnc": true}` or something to that affect, right?

steve-gregory:
Metadata will only be stored on the Servers DB. We won't add a new key/val to the Instance's metadata (Stored on the cloud). Additional servers can "add" the metadata via redeploy (Presuming you have the correct deployment key..)

> What's left for this to get merged?
This PR is waiting in the wings of #182. I'm creating the migrations now. This PR will be :shipit: shortly.

PR #181, opened 2016-08-04
fix "dict has no attr. 'name'" error

xuhang57:
changed `router.name` to `router['name'] `
changed `network.name` to `network['name']` 
in `service/networking.py`


steve-gregory:
Thanks! :shipit:

PR #182, opened 2016-08-09
Allocation Sources for Jetstream

amit4111989:
This PR contains all of the logic required for the MVP of Jetstream's Allocation Source feature.

Highlight Reel:

- [x] End user can select an allocationsource to charge on instance launch
- [x] End user can elect to change allocationsource for any given instance, during its lifecycle.
- [x] Reports are generated hourly and sent to the TAS API.

Post-Review ToDo:

- [x] DecimalField — 3 digits..
- [x] Global validation on events that "are not being handled by atmo"
- [x] _validate_ias_event possible to pass an allocation_source_id//instance_id that is not your own!
- [x] L192 fix before merge
- [x] resource_name and scheduler_id marked for deletion on TASAllocationReport?
- [x] ambiguous create_report task should be renamed
- [x] L46 of jetstream/tasks — Tacc_username (via driver.get_xsede_username(item.user.username) NOT item.user.username
- [x] Delete L58:L66 of jetstream/tasks.py
- [x] L85 — End_date timezone.now() may be bad here.. we want to explicitly set the end_date to be identical for all users TAS Reports — Also in total_usage()?
- [x] AllocationSourceSnapshot start_date should be "re-written each time"
- [x] EventTable change event user_burn_rate_changed to something that will allow compute_used + burn_rate
- [x] Create Report: indicate that the allocation_source parameter in this case is the name of the allocation_source.
- [x]  `instance_allocation_source_changed` Events: Update 'entity_id' to be Instance ID rather than username, include the username in the event payload BTS.

Post merge tests:
- [ ] Celery router doesn’t complain about missing tasks
- [ ] Digging into the speed of _get_user_burn_rate_snapshot (Low Priority)

Links to Troposphere PR: https://github.com/iPlantCollaborativeOpenSource/troposphere/pull/416


steve-gregory:
Review with @julianpistorius @amit4111989 

Atmobeta Checklist:
- Celery router doesn’t complain about missing tasks
- DecimalField — 3 digits..
- Search `title` is that right?
- Digging into the speed of _get_user_burn_rate_snapshot (Not duplicating calls)
- Global validation on events that "are not being handled by atmo"
- _validate_ias_event possible to pass an allocation_source_id//instance_id that is not your own!
- L192 fix before merge
- resource_name and scheduler_id marked for deletion on TASAllocationReport?
- ambiguous `create_report` task should be renamed
- L46 of jetstream/tasks — Tacc_username (via `driver.get_xsede_username(item.user.username)` *NOT* `item.user.username`
- Delete L58:L66 of jetstream/tasks.py
- L85 — End_date `timezone.now()` may be bad here.. we want to explicitly set the `end_date` to be identical *for all users TAS Reports* — Also in total_usage()?
- AllocationSourceSnapshot `start_date` should be "re-written each time"
- EventTable change event `user_burn_rate_changed` to something that will allow compute_used + burn_rate

PR #186, opened 2016-08-18
Updates to ExternalNetwork strategy

steve-gregory:
We made some incorrect assumptions about the external network strategy.

After some amount of clarification, these are the changes that were required in order to get ExternalNetwork support setup.

### DIFF
--- a/service/networking.py
+++ b/service/networking.py
@@ -112,6 +112,13 @@ class GenericNetworkTopology(object):
         """
         return True
 
+    def post_create_hook(self, network_driver, user_neutron, network_resources_dict):
+        """
+        Given the options in your strategy and your newly created resources,
+        use this space to "make the connections"
+        """
+        return
+
     def get_or_create_network(self, network_driver, user_neutron, network_name):
         network = network_driver.create_network(user_neutron, network_name)
         return network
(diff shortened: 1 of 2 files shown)


service/networking.py
```diff
@@ -229,19 +236,10 @@ def __init__(self, identity):
...
-
-    def get_or_create_network(self, network_driver, user_neutron, network_name):
-        # Step 1. Does public network exist?
+    def get_public_network(self, network_driver):
+        """
+        This method is special to ExternalNetwork
+        """
         public_network = network_driver.find_network(
             self.external_network_name)
         if type(public_network) == list:
```
cdosborn: What if public_network is an empty list?

steve-gregory: validate() call should raise exception before you get here

cdosborn:
ill tack on a commit here, its easier sometimes to show code.

cdosborn:
I did notice that the `gateway` variable was not being used.

steve-gregory:
@cdosborn Thanks! This is where I was intending to go after I had commited the "parity" version, so we're both on the same page

steve-gregory:
TODO:

- [ ] Test this version of moc-master and be sure it doesn't break networking

cdosborn:
i'll buy you mcdees if my code works.

cdosborn:
@o@ IT WORKED???

PR #189, opened 2016-08-23
Create a setting to blacklist tags

steve-gregory:
Rather than introduce more database migrations, this is a simple approach that will allow the Troposphere UI to block users from submitting tags that are reserved for staff users *only*.

- [x] Allow Troposphere to provide user with a clean list ahead of time.
- [x] Create a new 'field' for the Tag API endpoint (allow_access) that is True if that user can add the tag to their image.
- [x] Atmosphere should also filter and remove tags that are in the BLACKLIST_TAGS passed directly to the MachineRequest API.

[Depends-on-Troposphere PR](https://github.com/iPlantCollaborativeOpenSource/troposphere/pull/435)

### DIFF
--- a/api/v2/views/machine_request.py
+++ b/api/v2/views/machine_request.py
@@ -13,8 +13,10 @@ from atmosphere import settings
 
 from core import exceptions as core_exceptions
 from core.email import send_denied_resource_email
-from core.models import MachineRequest, IdentityMembership, AtmosphereUser,\
-    Provider, ProviderMachine, Group
+from core.models import (
+    MachineRequest, IdentityMembership, AtmosphereUser,
+    Provider, ProviderMachine, Group, Tag
+)
 from core.models.status_type import StatusType
 from core.email import requestImaging
 
@@ -40,10 +42,10 @@ class MachineRequestViewSet(BaseRequestViewSet):
                     Q(start_date__gt=timezone.now() - timedelta(days=7))
                 )
             )
-            if self.request.user.is_staff:
+            if request_user.is_staff:
                 return all_active.order_by('-start_date')
             return all_active.filter(
-                created_by=self.request.user
+                created_by=request_user
                 ).order_by('-start_date')
         return super(MachineRequestViewSet, self).get_queryset()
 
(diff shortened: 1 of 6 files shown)


steve-gregory:
This PR is ready for review 👓

api/v2/views/machine_request.py
```diff
@@ -40,10 +42,10 @@ def get_queryset(self):
                     Q(start_date__gt=timezone.now() - timedelta(days=7))
                 )
             )
-            if self.request.user.is_staff:
+            if request_user.is_staff:
```
lenards: Is `request_user` injects by DRF? I thought it was `request.user` that was the associated user object.

api/v2/views/machine_request.py
```diff
@@ -62,11 +64,23 @@ def _get_new_provider(self):
...
+            if not core_tag:
+                continue
+            if not core_tag.allow_access(user):
+                continue
+            new_tag_names.append(tag)
+        return ", ".join(new_tag_names)
+
     def perform_create(self, serializer):
 
+        request_user = self.request.user
```
lenards: Wouldn't this need to be an instance field to be available in the `get_queryset()`? Or magic ✨ ?

api/v2/views/machine_request.py
lenards: Maybe my DRF ignorance is showing ... or maybe (**gasp**) my Python ignorance!

lenards: I chatted with Julian and PyCharm seems to be confirm that I am right that `request_user` is not a known system. Not sure if there is something missing here.

lenards: Naïvely, I think this would be fixed with: 
```python
    request_user = self.request.user
   if request_user.is_staff:
   # ... 
```

steve-gregory: You right you right. `:%s/self.request.user/request_user` you have failed me 😄

lenards: 👍

lenards:
I think this looks good now.

lenards:
:wand:

steve-gregory:
:bullettrain_front: :shipit:

PR #191, opened 2016-08-28
Problem: Failing tests

julianpistorius:
Fixed tests. See commits.

### DIFF
--- a/jetstream/tests.py
+++ b/jetstream/tests.py
@@ -3,6 +3,7 @@ from urlparse import urlparse
 
 import vcr
 from django.test import TestCase
+from mock import Mock
 
 
 def scrub_host_name(request):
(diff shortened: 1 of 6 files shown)


jetstream/tests.py
```diff
@@ -29,20 +30,33 @@ class TestJetstream(TestCase):
...
         pass
 
+    def _check_cassette(self, cassette, expected_cassette_length):
+        """Make sure the cassette is:
+         1. The expected length and either
+         2. New (dirty) or
+         3. Existing, rewound
+         """
+        self.assertEqual(len(cassette), expected_cassette_length)
+        self.assertTrue(cassette.dirty or cassette.all_played)
```
cdosborn: I have two nitpicks about this method. I would name it `assert_cassette_playback_length`. When I first saw `check_cassette` i had to go look it up, because it wasn't clear (why does it require a number as an argument?). The other is that this method should exist outside of this test. It could also be a decorator, since it is something we would apply to most cassette tests. 

major 👍  for working on tests.

run_tests.sh
```diff
@@ -3,4 +3,4 @@ set -e
 VIRTUALENV=/opt/env/atmo
 SCRIPT_DIR=$(dirname `readlink -f $0`)
 source $VIRTUALENV/bin/activate
-$SCRIPT_DIR/manage.py test --noinput --settings=atmosphere.settings -v2 -- $@
+$SCRIPT_DIR/manage.py test --keepdb --noinput --settings=atmosphere.settings -v2 -- $@
```
cdosborn: 👍

jetstream/tests.py
julianpistorius: Thanks @cdosborn. That's a good idea!

julianpistorius: I couldn't immediately figure out how to use it as a decorator. Open to suggestions.

cdosborn:
Should this be re-targeted at master?

julianpistorius:
@cdosborn: Hmm... probably. @steve-gregory?

julianpistorius:
Superseded by: PR https://github.com/iPlantCollaborativeOpenSource/atmosphere/pull/194

PR #192, opened 2016-08-29
Enforce allocation sources

steve-gregory:
- Disables 'monitor_instance_allocations'
- New hook listens on AllocationSourceSnapshot. If the compute is >100%, enforce allocations (once, and note it in the event, and ignore it for all future calls where compute is >100%)
- Run the enforcement task Asynchronously so that users are not "stuck inside the hook"
- Move from `pre_save` to `post_save` for `listen_before_allocation_snapshot_changes`


steve-gregory:
Merging due to time limitations. Using this PR to keep track of the history.

PR #197, opened 2016-09-08
Problem: On Jetstream production server allocation usage calculation …

julianpistorius:
…is wrong

Solution: Specify an explicit ordering when querying instance_status_history.

It turns out that on the Jetstream server these are returned in reverse chronological order by default. This caused the calculation to be completely wrong.


steve-gregory:
:shipit:

PR #198, opened 2016-09-09
Add subnet pool specification with a default of None

jmlowe:
Starting with the kilo release neutron can use a shared subnet pool,
this ensures unique subnets without the need to probe.  When a subnet
pool id is configured skip the search and CIDR specification and just
create the subnet using the subnet pool.

This may need some further polish, I took the per provider override for settings at face value.  One possible improvement would be to use subnet pool name and lookup the id.


steve-gregory:
Conflicts can't be resolved under this PR -- see https://github.com/cyverse/atmosphere/pull/258

PR #199, opened 2016-09-19
No spaces allowed in list of usernames

c-mart:


steve-gregory:
Looks great, thanks @c-mart

PR #200, opened 2016-09-20
Fix instance hung indefinitely in active deploy

cdosborn:
A user's instance was stuck in "active-deploying". After several re-deploys, I looked into the culprit. The instance deploy task will retry until it fails. Once it fails we call another task to update the instance with status "error-deploy error". That second task was never called.

It turns out that if the time limit is exceeded (while it's retrying) the first task is terminated and the second is never run. By using a soft time limit, the second will get a chance to do cleanup, if a timeout occurs.

I verified that this was indeed the culprit by testing very small time_limits and small soft_time_limits to compare.

### DIFF
--- a/service/tasks/driver.py
+++ b/service/tasks/driver.py
@@ -1089,7 +1089,7 @@ def _deploy_instance_for_user(driverCls, provider, identity, instance_id,
 
 @task(name="_deploy_instance",
       default_retry_delay=124,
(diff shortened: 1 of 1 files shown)


steve-gregory:
> cdosborn changed the base branch to iPlantCollaborativeOpenSource:ravenous-raven from iPlantCollaborativeOpenSource:master 11 minutes ago

AMAZED 🐶 :doge: :shipit:

PR #201, opened 2016-09-22
Problem: Jetstream and EventTable tests fail with default settings

julianpistorius:
Solution: A. Skip Jetstream tests if the 'jetstream' app is not enabled, and B. override the allocation source warning thresholds to work with the EventTable test logic.


steve-gregory:
👓 :shipit: Thanks @julianpistorius

PR #202, opened 2016-09-23
Fix requirements hook falsy test

cdosborn:
Only `nil` and only `false` are falsy in ruby. I made the mistake of thinking the empty string was falsy.

Now the hook should only speak up when it's supposed to!

### DIFF
--- a/contrib/requirements-warning.rb
+++ b/contrib/requirements-warning.rb
@@ -1,24 +1,19 @@
-DEBUG = false
-
 # Get the commits before the merge
 #   ex. heads = ["<latest commit of master>", "<previous commit of master>", ...]
 heads = `git reflog -n 2 | awk '{ print $1 }'`.split
 
 # Make sure our revision history has at least 2 entries
 if heads.length < 2
-    exit 0 
+    exit 0
 end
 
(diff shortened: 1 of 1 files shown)


contrib/requirements-warning.rb
```diff
@@ -1,24 +1,18 @@
...
-    red = "\e[31m" 
+    red = "\e[31m"
     puts "[#{red}requirements-warning.rb hook#{default}]: New python requirements."
-    puts "pip install #{diffFiles}"
-
-    if DEBUG then
-        # Print the diff with 0 lines of context
-        puts `git diff -U0 #{heads[1]} #{heads[0]} -- #{diffFiles}`
-    end
+    puts "pip install *requirements.txt"
```
cdosborn: shaddup robot

contrib/requirements-warning.rb
```diff
@@ -1,24 +1,18 @@
-DEBUG = false
-
 # Get the commits before the merge
 #   ex. heads = ["<latest commit of master>", "<previous commit of master>", ...]
 heads = `git reflog -n 2 | awk '{ print $1 }'`.split
 
 # Make sure our revision history has at least 2 entries
 if heads.length < 2
-    exit 0 
+    exit 0
```
cdosborn: shaddup

PR #209, opened 2016-10-05
Project sharing

steve-gregory:
**WIP**
This is the work done so far in the way of project sharing. The description of this PR will change as the feature nears completion.

**Highlight Reel**:
- [x] Create a new db migration and new core Model `ProjectMembership` between `Project` and `Group`[1] and an additional attribute `can_edit` that is False by default. 
- [x] Create API scaffolding
- [x] Create core.query methods to encapsulate the logic required for 'simple statements'
- [x] Create a set of permissions using core.query to allow/dis-allow shared users based on their privileges (or if they themselves were the creator) `ProjectMembers` that are created by default with  `can_edit=False` will receive a 403 with the message `"You must be the project owner to execute this command."`
- [x] Apply the set of permissions to all of the project resources: Instance, Volume, ExternalLink, Application(a.k.a. Image)
- [x] Apply the set of permissions to APIs related to ProjectResource changes (Deleting/moving from a project)
- [x] If you have the ProjectOwner permission, you should be able to _act on_ (Reboot, Suspend, Resume, etc.) resources in the project.
- [x] If you have the ProjectOwner permission, you should be able to edit details on resources in the project.
- [ ] If you have the ProjectOwner permission, you should be able to launch/create resources "On behalf of the project"[2] 

**Not Included**
This PR will not address Image Sharing nor the maintenance of editing details on "Images".

_References_:
[1] - Traditionally, all *Membership classes are pointed at the Group rather than the AtmosphereUser to allow for future flexibility
[2] - This is a _loose definition_ because identities are not _directly_ tied into projects. Users create ephemeral projects, but the identities they have stick with them. Need to decide what we mean here.


arleigh16:
unsubscribe

steve-gregory:
Hey @arleigh16 !

Looks like you are trying to disable notifications for Atmosphere -- But sending an email won't do anything for you!

 Instead, navigate here:
https://github.com/iPlantCollaborativeOpenSource/atmosphere

Then select `Unstar`, `Unwatch` and/or `Unfollow` and you should be able to remove yourself.

Thanks,
Steve Gregory

lenards:
✅ from the :dog2:

steve-gregory:
After looking at this PR in further detail, I have decided that it will never be more than a half-baked feature. The alternative method, #226, will be pursued instead.

PR #210, opened 2016-10-05
Create an Expiration Plugin for LDAP

steve-gregory:
**Highlight Reel:**
- [x] Create a new PluginManager class to avoid DRY-violation in plugins.py
- [x] Create a new class to test if an LDAP user has an expired password
- [x] Add `is_expired` to the API `profile` endpoint


steve-gregory:
Merging this into Master to gain access to the PluginManager.

This PR has no net affect until Troposphere decides to act on `user_profile.is_expired == True`..

PR #213, opened 2016-10-18
Atmo 1540

amit4111989:
Fixes the allocation logic by taking into account both payloads for instance allocation change event


api/v2/serializers/details/event.py
```diff
@@ -11,14 +11,13 @@ class EventSerializer(serializers.HyperlinkedModelSerializer):
     def _validate_ias_event(self, entity_id, payload):
         allocation_source_id = payload.get('allocation_source_id', '')
         instance_id = payload.get('instance_id', '')
-        if entity_id != instance_id:
+        user = self._get_request_user()
+        if entity_id != user.username:
             raise serializers.ValidationError(
-                "Expected entity ID to be the Instance ID: %s, Received: %s"
+                "Expected entity ID to be the Username: %s, Received: %s"
                 % (instance_id, entity_id))
```
steve-gregory: This error should be printing out `user.username` rather than `instance_id`

service/allocation_logic.py
```diff
@@ -26,7 +26,8 @@ def create_report(report_start_date, report_end_date, user_id=None, allocation_s
         else:
             return data
     else:
-        write_csv(data)
+        return data
+        #write_csv(data)
```
steve-gregory: Toss this line? I think if we intend to keep `write_csv` around we should make it a new kwarg for `create_report()`.

steve-gregory:
:shipit:

PR #214, opened 2016-10-18
Fix API to include 'initializing' when calling 'restore IP chain'

steve-gregory:
Update the 'restore_ip_chain' to set the instance to 'Initializing' *before* any other actions take place to avoid _bounce-back_ from the API

Troposphere PR: https://github.com/iPlantCollaborativeOpenSource/troposphere/pull/484


steve-gregory:
Self-:shipit: #noshame

PR #215, opened 2016-10-26
Fix search 🎉 

cdosborn:
There was an exponential slowdown when a user searched for images. Without getting into the details `filter(testA).filter(testB).filter(testC)` can result in very very bad sql. django-rest-framework was doing this and we were as well. The solution is to use `filter(testA & testB & testC)`. Rather than wait for drf to fix this issue, we have monkey patched the troublesome method (when they release a fix, we'll revert the patch).

Read a ticket I filed against Django:
https://code.djangoproject.com/ticket/27388

Here is a general thread:
https://code.djangoproject.com/ticket/27303

Here is a rough estimate of the performance for the underlying sql, based on the number of terms in a search.

```
# terms    time(ms)
===================
1          0.047
2          0.056
3          0.066
4          0.076
5          0.093
6          0.108
7          0.117
8          0.124
9          0.140
10         0.149
```

For perspective, 2 terms used to take ~36s on my dev box.

### DIFF
--- a/api/v2/views/image.py
+++ b/api/v2/views/image.py
@@ -9,6 +9,39 @@ from api.v2.views.mixins import MultipleFieldLookup
 from core.models import Application as Image
 
 
+#
+# The following imports and method and monkey patch are a quick fix for a big
+# problem.  The patch should be removed when the equivalent drf method does
+# not chain filter methods. Chaining filter methods with m2m relations,
+# can result in /very/ bad performance. For more information see:
+#
+#   https://code.djangoproject.com/ticket/27303#comment:26
+#
+from django.utils import six; from django.db import models; import operator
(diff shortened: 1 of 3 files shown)


julianpistorius:
Whoot! 🍾 🏁

lenards:
Awesome work Connor!

julianpistorius:
👍 Looks good to me @cdosborn!

Minor tweak & some basic sanity tests: https://github.com/cdosborn/atmosphere/pull/3

cdosborn:
I brought in @julianpistorius's changes. The test logic will require more work, before it can be included.

I also added more comments around the queries.

steve-gregory:
:shipit: Thanks @cdosborn !!

PR #217, opened 2016-10-31
Reporting tests and tweaks.

julianpistorius:


api/v2/views/reporting.py
```diff
@@ -75,29 +70,37 @@ def create_excel_file(self, raw_dataframe, writer):
         if len(raw_dataframe.index) <= 1:
             return None
         frequency = self.set_frequency()
-        new_datasets = self._create_datasets(raw_dataframe, writer, frequency)
+        new_datasets = self._create_datasets(raw_dataframe, frequency)
```
julianpistorius: Unused parameter?

api/v2/views/reporting.py
```diff
@@ -75,29 +70,37 @@ def create_excel_file(self, raw_dataframe, writer):
...
         writer = self._format_and_print_workbook(writer, new_datasets, frequency)
         return writer
 
-    def _create_datasets(self, raw_dataframe, writer, frequency):
+    @staticmethod
```
julianpistorius: A few methods which can be static - or potentially just functions.

api/v2/views/reporting.py
```diff
@@ -207,13 +217,16 @@ def get_queryset(self):
             instances_qs = Instance.objects.all()
         elif request_user.is_authenticated():
             instances_qs = Instance.for_user(request_user)
+        else:
+            raise exceptions.NotAuthenticated()
```
julianpistorius: Redundant, but I'm paranoid.

api/v2/views/reporting.py
```diff
@@ -242,5 +255,11 @@ def list(self, request, *args, **kwargs):
...
         query_params = self.request.query_params
         if not query_params.items():
-            return failure_response(status.HTTP_400_BAD_REQUEST, "The reporting API should be accessed via the query parameters: ['start_date', 'end_date', 'provider_id']")
-        return super(ReportingViewSet, self).list(request, *args, **kwargs)
+            return failure_response(status.HTTP_400_BAD_REQUEST,
+                                    "The reporting API should be accessed via the query parameters:"
+                                    " ['start_date', 'end_date', 'provider_id']")
+        try:
+            results = super(ReportingViewSet, self).list(request, *args, **kwargs)
+        except ValueError:
```
steve-gregory: Whats this for?

api/tests/v2/test_reporting.py
```diff
@@ -0,0 +1,106 @@
...
+from test_utils.comparison_utils import dict_eq_
+
+
+class ReportingTests(APITestCase):
+    def setUp(self):
+        self.anonymous_user = AnonymousUserFactory()
+        self.user = UserFactory.create()
+        self.view = ReportingViewSet.as_view({'get': 'list'})
+
+    def test_a_sanity_check(self):
```
steve-gregory: Haii Thanks!

api/v2/views/reporting.py
julianpistorius: In retrospect there is a better way to do this We could add a try-catch for ValueError in the [`get_filter_query` method](https://github.com/iPlantCollaborativeOpenSource/atmosphere/pull/216/files#diff-f96940a27701dfcc2c1602d0fb5b7cdfR229), then raise a custom exception to catch here.

Or even better is what you suggested - which is to just use DRF's query parameter validation/serialization. :)

PR #218, opened 2016-10-31
Problem: Atmosphere Django tests are slow

julianpistorius:
Solution: Disable database serialization. We don't need it, and it slows things down. See: https://docs.djangoproject.com/en/1.10/ref/settings/#serialize


steve-gregory:
Good find @julianpistorius 🔎

PR #219, opened 2016-11-02
Problem: Allocation calculation takes long - almost 100% stuck in Pos…

julianpistorius:
…tgreSQL

Solution: Add indices to `event_table`, including a JSON index on `payload`.
Note: This is a quick, low-risk fix while I integrate the new stream-based solution.


steve-gregory:
:shipit: thanks @julianpistorius

PR #220, opened 2016-11-03
Bugfix: Image catalog will now skip private images not created...

steve-gregory:
... by tenants/projects who have listed their image as 'private' and are not atmosphere users.

Additionally, we can skip 'public' images if they were created *outside* the domain managed by atmosphere. 

Ideally, I would like to 'flag' support for this on/off depending on operator preference.. But for now its on by default...


julianpistorius:
👍 Looks like it should do the job.

PR #223, opened 2016-11-07
Problem: Jetstream allocation source usage appears too high

julianpistorius:
Solution: We should be treating TAS as the source of truth, instead of calculating it ourselves.

- Integrated and refactored @amit4111989's code for using the TAS API as source of truth
- I have some offline tests with production data which give me confidence in these new calculations, but I would still like to create tests using test data
- Modified signature of `update_snapshot` so I can pass in start & end dates

TODO:
- Corrective reports
- Tests with clean test data

### DIFF
--- a/jetstream/tasks.py
+++ b/jetstream/tasks.py
@@ -12,6 +12,8 @@ from core.models.allocation_source import (
     AllocationSource, UserAllocationSnapshot
 )
 from core.models.event_table import EventTable
+from service.allocation_logic import create_report
+from core.models.user import AtmosphereUser
 
 from .models import TASAllocationReport
 from .allocation import (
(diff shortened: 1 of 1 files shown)


jetstream/tasks.py
```diff
@@ -137,35 +132,67 @@ def send_reports():
...
+            allocation_source_id=allocation_source_id,
+            user_id=relevant_users[username],
+            defaults={
+                'compute_used': round(compute_used / 3600, 2),
+                'burn_rate': burn_rate
+            }
+        )
+        allocation_source_burn_rates[allocation_source_name] += burn_rate
+
+    tas_api_obj = TASAPIDriver()
```
steve-gregory: `update_snapshot()` takes a start_date/end_date args that are not used in the code below this line..

It might be best if we separate the concerns here by keeping this code in a different function.

jetstream/tasks.py
```diff
@@ -137,35 +132,67 @@ def send_reports():
...
-            logger.info("Total usage for AllocationSource %s (%s-%s) = %s (Burn Rate: %s)" % (source.name, start_date, end_date, compute_used_total, global_burn_rate))
-        payload_as = { 
-            "allocation_source_id":source.source_id, 
-            "compute_used":compute_used_total,
-            "global_burn_rate":global_burn_rate
-        }
-        EventTable.create_event("allocation_source_snapshot", payload_as,source.name)
+    end_date = end_date or timezone.now()
+    # TODO: Read this start_date from last 'reset event' for each allocation source
+    start_date = start_date or '2016-09-01 00:00:00.0-05'
```
steve-gregory: 👍

jetstream/tasks.py
```diff
@@ -137,35 +132,67 @@ def send_reports():
...
-            # determine end date and start date using last snapshot
-            start_date = user.date_joined
-            # calculate compute used and burn rate for the user and allocation source combo
-            compute_used, burn_rate = total_usage(user.username,start_date,allocation_source_name=source.name,end_date=end_date,burn_rate=True)
-
-            allocation_source_total_compute[source.name] = allocation_source_total_compute.get(source.name,0) + compute_used
-            allocation_source_total_burn_rate[source.name] = allocation_source_total_burn_rate.get(source.name,0) + burn_rate
-
-            payload_ubr = {"allocation_source_id":source.source_id, "username":user.username, "burn_rate":burn_rate, "compute_used":compute_used}
-            EventTable.create_event("user_allocation_snapshot_changed", payload_ubr, user.username)
```
steve-gregory: This event was removed and is not re-added via hook or other means. Is the intent here to remove this event from being created (to avoid table bloat)?

jetstream/tasks.py
```diff
@@ -137,35 +132,67 @@ def send_reports():
...
-        compute_used_total = allocation_source_total_compute.get(source.name,0)
-        global_burn_rate = allocation_source_total_burn_rate.get(source.name,0)
-        if compute_used_total != 0:
-            logger.info("Total usage for AllocationSource %s (%s-%s) = %s (Burn Rate: %s)" % (source.name, start_date, end_date, compute_used_total, global_burn_rate))
-        payload_as = { 
-            "allocation_source_id":source.source_id, 
-            "compute_used":compute_used_total,
-            "global_burn_rate":global_burn_rate
-        }
-        EventTable.create_event("allocation_source_snapshot", payload_as,source.name)
```
steve-gregory: Same question above, here..

jetstream/tasks.py
julianpistorius: Correct!

jetstream/tasks.py
julianpistorius: Ditto.

jetstream/tasks.py
julianpistorius: Weird. I did have something that uses it in there. Mainly I used it when running tests.

steve-gregory:
After speaking with @julianpistorius the function (which is working in its current implementation) will be merged as-is with the goal of splitting it out into two functions in a future release.

Additionally, with the removal of `allocation_source_snapshot` event, we will table making a decision until we have a solution for (Allocation Source + CyVerse) and can ensure that the question of "Allocation Source change over time" can be effectively answered. If not, this event _might_ make a resurgence.

PR #224, opened 2016-11-07
Use the same logic to skip machines in 'monitor_machines' and 'prune_machines'

steve-gregory:


steve-gregory:
After chatting with @julianpistorius it might be better to do the following (in addition to re-target to the `toco-toucan`/`master` branch)

- `prune_machines` as run by celery should always perform a 'dry-run' and return the "list of images that _would_ be pruned".
  - The list could potentially be e-mailed to the ADMINS for further review.
- Create an 'action', to be run by administrators, that will allow this method to be performed as appropriate.

PR #225, opened 2016-11-07
Faster update snapshot hotfix burn rate

julianpistorius:
Problem: Jetstream allocation source burn rate appears high

Solution: Re-calculate burn rate per user/allocation/instance.

I misunderstood the burn rate column in the report rows. It was returning the cumulative burn rate. I added a separate calculation.


steve-gregory:
:shipit:

PR #226, opened 2016-11-08
Project Sharing via User Group Mapping

steve-gregory:
Required to view changes: https://github.com/cyverse/troposphere/pull/570

In this method of Project sharing, we take an existing 'structure' from openstack (User-> tenant) and provision atmosphere with the same structure (User -> Group).

For example, if Openstack structure looks like:
```
User:sgregory Tenant:atmosphere-devs
User:lenards Tenant:atmosphere-devs
User:edwins Tenant:atmosphere-devs
``` 
Atmosphere would create:
```
AtmosphereUser:sgregory Group:atmosphere-devs
AtmosphereUser:lenards Group:atmosphere-devs
AtmosphereUser:edwins Group:atmosphere-devs
```
In addition to creating the identity, credentials, etc.

In order to facilitate this change (and still do things "a different way") I have created an "Account creation plugin". The default plugin for cyverse is to create `AtmosphereUser: sgregory Group:sgregory` during identity creation.

AccountCreation-as-a-plugin Highlight Reel:
- [x] Include deletion to 'reverse the flow'
- [x] Create a Plugin for the creation of accounts via CyVerse+Grouper
- [x] Create a Plugin for the creation of accounts via Jetstream+TACC 
- [x] Create a new PluginManager (Single string rather than list)
- [x] Adjust 'create_usergroup()' to take group_name and 'is_leader'
- [x] Adjust 'create_identity' to take args above + 'account_user'
- [x] Create a unique user on a unique group, with a unique openstack
username and a unique project_name
	new file:   atmosphere/plugins/accounts/__init__.py
	new file:   atmosphere/plugins/accounts/creation.py
	modified:   core/plugins.py
- [x] determine how to best query 'shared_identities' for (Leadership True/False/Both), then create methods to return the resources in a similar fashion (Instance,Volume, etc.)
- [x] Leader-required permissions enforced API level (v1 and v2)
- [x] Null Project Resources should more-closely mirror behavior of "Null allocation source modal" to allow *each resource to have a selection-value*.
- [x] Project-specific filter *required* for null resources to ensure they are placed in the right bin.

TODO:
- [ ] Profile calls are taking too long, need to be able to bail-out when identities have been found.


steve-gregory:
Rebased to `x-x` - New PRs coming soon.

PR #229, opened 2016-11-28
Change old reference from quota.identitymembership to quota.identity

cdosborn:
Creating a volume threw an api error because it would check a user's quota.
This fixes get_quota for an identity.


cdosborn:
🔥 🚑 🔥

PR #230, opened 2016-11-28
Prefer provider location rather than uuid in support emails

cdosborn:
When I deal with a ticket its not helpful for me to have the provider uuid, i need to know at a glance the location (Tucson vs Workshop). 

@steve-gregory Do you ever need the uuid?


steve-gregory:
The location is not guaranteed the be unique, perhaps we want to make it unique as well so that it can be used as an identifier?

cdosborn:
I was wondering if you ever use the provider uuid from the email, if so I can add in both location and uuid.

PR #231, opened 2016-11-30
Fix/refactor incorrect equality test with default allocation

cdosborn:


steve-gregory:
@cdosborn if this is indeed a bug, should we move this into the U release?

cdosborn:
@steve-gregory i don't think its high pri, the incorrect logic meant that people with base allocation were getting reset to the base allocation unnecessarily when we moved to making allocations non-unique, not a big deal

PR #232, opened 2017-01-03
Fix disappearing image bookmarks on refresh

cdosborn:
The bookmark queryset was incorrect it had a filter that was returning an
empty list.


lenards:
This is a fix for [ATMO-1604](https://pods.iplantcollaborative.org/jira/browse/ATMO-1604).

lenards:
Connor explained the logic to me. I think he's got the correct characterization in place. 

We want to ensure that if an image has an end-date, and it is in the _future_, then we allow it to be bookmarked. So, it seems that the filter should be: 

_"end date is null or end date is in the future"_

I believe this maintains the expected business intent.

lenards:
Or, 👍 :shipit:

steve-gregory:
@cdosborn thanks for the fix!

PR #234, opened 2017-01-05
Fix proxy pass so ngnix server starts

cdosborn:
Nginx was failing, it's configtest was not passing. The reason was because the
server specified in the proxy_pass directive was not resolving. Nginx checks
these beforehand to help. Nginx cannot resolve SERVER_URL (i.e.
app.atmo.cloud) so it fails. The solution is to use localhost.

### DIFF
--- a/extras/nginx/Makefile.j2
+++ b/extras/nginx/Makefile.j2
@@ -23,26 +23,20 @@ JENKINS_CONF_FILE=jenkins.conf
 JENKINS_CONF_PATH=$(ATMOSPHERE_PATH)/extras/nginx/locations/$(JENKINS_CONF_FILE)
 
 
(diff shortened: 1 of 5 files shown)


extras/nginx/Makefile.j2
```diff
@@ -23,27 +23,21 @@ JENKINS_CONF_FILE=jenkins.conf
...
 
 setup-atmo:
 	$(VIRTUALENV_PATH)/bin/python $(ATMOSPHERE_PATH)/configure -c nginx:nginx-atmo
 
-setup-flower:
-	$(VIRTUALENV_PATH)/bin/python $(ATMOSPHERE_PATH)/configure -c nginx:nginx-flower
-
-setup-jenkins:
-	$(VIRTUALENV_PATH)/bin/python $(ATMOSPHERE_PATH)/configure -c nginx:nginx-jenkins
-
```
steve-gregory: If we're going to drop these lines, and this solution works for local developers *and* standalone servers, we should probably drop the two .j2 files

cdosborn: :man_facepalming: oops, thought i had

cdosborn: If you look at the commits, files were renamed from .conf.j2 -> .conf

PR #237, opened 2017-01-11
Access atmosphere by Token

steve-gregory:
This PR introduces the ability to access cloud resources **without saving the users password in the credentials.**

In order to do this, we rely on https://github.com/cyverse/django-cyverse-auth/pull/6 to connect to **a single openstack provider** and use keystone as the Identity Provider. If authorized, keystone will return an unscoped token that Atmosphere can be saved to a user's identity (via 
https://github.com/cyverse/troposphere/pull/524).

This PR removes a large portion of dependencies on the account driver, but I was not able to completely remove account drivers (which require a provider with access to an admin-role identity) from areas related to monitoring machines, sizes, and instances across the userbase.

Requires the latest revision of rtwo(https://github.com/cyverse/rtwo/pull/6) in order to avoid adjusting or including the auth_url to create network drivers.


xuhang57:
Thanks for the great work. Just one thing that I noticed, we are adding project_name along with username and password as the variables needed when a user login. So, does it mean a user needs to type the project name also during login to Atmosphere? 

Thanks!

atmosphere/settings/local.py.j2
```diff
@@ -303,6 +307,9 @@ AUTHENTICATION_BACKENDS = (
     # For Service-Access
     'django_cyverse_auth.authBackends.LDAPLoginBackend',
     {% endif %}
+    {% if AUTH_ENABLE_OPENSTACK -%}
+    'django_cyverse_auth.authBackends.OpenstackLoginBackend',
+    {% endif -%}
```
steve-gregory: Reminder: Documenting the new variables, what is expected and ensuring "the secrets repos" have the updated variables pre-merge.

core/models/quota.py
```diff
@@ -235,7 +235,13 @@ def has_port_count_quota(driver, quota, new_size=0, raise_exc=True):
     # Always True if port_count is null
     if not quota.port_count or quota.port_count < 0:
         return True
-    fixed_ips = [port for port in driver._connection.neutron_list_ports() if port['device_owner'] == 'compute:None']
+    # Consider it true if we fail to connect here
+    try:
+        port_list = driver._connection.neutron_list_ports()
```
steve-gregory: Reminder: This may need to be fixed because its a `neutron_` method

variables.ini.dist
```diff
@@ -59,6 +60,8 @@ GLOBUS_TOKEN_URL = ; https://globus.server.org/token
 GLOBUS_AUTH_URL = ; https://globus.server.org/authorize
 OAUTH_CLIENT_KEY = ; oauth_key
 OAUTH_CLIENT_SECRET = ; oauth_secret
+KEYSTONE_SERVER = ; "https://my-openstack.cloud.org:5000"
+KEYSTONE_DOMAIN_NAME = ; "default"
```
steve-gregory: Suggestion: Including explanation on when/how/why to use variables could be done in clank (or wherever we are including docs)

steve-gregory:
@julianpistorius mentioned that before merging I should include tests for:
```
API test for auth
API test for token_update
(Troposphere) API test for login
```

variables.ini.dist
steve-gregory: See https://github.com/cyverse/clank/pull/153 for more thorough description.

lenards:
👏

PR #240, opened 2017-01-18
This script will print instances in networking or deploy_error

cdosborn:
@steve-gregory the instructions to run this script could be improved, what do you suggest?


steve-gregory:
My suggestion is: argparse all-the-things (See https://github.com/cdosborn/atmosphere/blob/2271a470258679c7d6202eb82d48714c1cfc3a6d/scripts/add_new_accounts.py#L26-L44)

cdosborn:
@steve-gregory I wrapped it in argparse. Also now it catches when your environment isn't setup.

steve-gregory:
:shipit: Thank you @cdosborn

PR #241, opened 2017-01-21
Problem: Dependencies can change

julianpistorius:
If we don't lock them down we can't have repeatable or reliable builds.

Solution: Use pip-tools to lock them down.

See https://github.com/nvie/pip-tools  and http://nvie.com/posts/better-package-management/ for more information.

Note: I've also split the dev & prod requirements files.

Hint: Use the `pip-sync` command to manage the state of a virtual environment. It makes all the packages in a virtual environment conform to exactly the versions specified in the requirements file(s). It will uninstall packages which are not in the requirements files. This means that you don't have to delete a virtual environment and recreate it to be sure the correct packages are installed. It also means you can switch back and forth between different branches of a project and be confident that the correct packages for the branch is installed. For example:

```bash
source /opt/env/atmo/bin/activate
pip install pip-tools
pip-sync dev_requirements.txt requirements.txt
# All packages with correct versions in both requirements files should now be installed.
pip-sync requirements.txt
# Only packages (and versions) specified in requirements.txt will now exist in the virtual environment.
```


steve-gregory:
Thanks for setting this up! :shipit:

julianpistorius:
@steve-gregory: How do we want to get this into voracious-velociraptor?

PR #242, opened 2017-01-24
Fix pip tools

julianpistorius:


steve-gregory:
Thanks @julianpistorius :shipit:

PR #243, opened 2017-01-24
Removing lots of code that I suspect is dead

c-mart:
> [16:16]  cmart
Steve, Connor and I are wondering what all this is (still) doing here
https://github.com/cyverse/atmosphere/tree/master/init_files
there's some Ruby code and we are scared
[16:17]  sgregory
Burn it to the ground
Nuke it from orbit
Pre Ansible and pre deployment
The era of cloud init and "atmo boot" *shudders*

So I did, and followed the trail of code that refers to things in `init_files/` which I suspect are also no longer used. Also removed some imports that PyCharm says are unused.

To do:

- [x]  Ensure that Atmosphere still works, e.g. can launch instances, etc

Tada:

- [x] Ensure that Clank completes run without error


init_files/boot_scripts/hello_world.sh
```diff
@@ -1,10 +0,0 @@
-#!/bin/bash
-
-################################################################
-#
-#  This Script can say 'Hello World'
-#  
-#  hello_world.sh -- Version 1.0
-################################################################
-
-echo 'Hello World. This is an example of a boot script!'
```
steve-gregory: These files might be nice to have in an "example boot scripts" directory, which is what they were intended for

init_files/v1/iplant_backup.sh
```diff
@@ -1,346 +0,0 @@
-#!/bin/bash
-
-################################################################
-#
-#  Script to backup or retrieve data between an Atmosphere
-#  instance and iPlant Datastore
-#  
-#  iplant_backup.sh -- Version 1.0
```
steve-gregory: Is this iplant_backup script stored elsewhere? I believe that this was being used by end-users at one point in time...

c-mart:
Tested this PR in an atmo-local Vagrant environment. Details of test environment follow:
- Connected to Jetstream test cloud using [Connecting a Cloud Provider](https://cyverse.github.io/atmosphere-guides/connecting_cloud_provider.html) guide
- Had to switch Troposphere from master to voracious-velociraptor
- Had to merge https://github.com/cyverse/atmosphere-ansible/pull/59 in order for Ansible to complete anything
- Then, had to use atmosphere-ansible from jetstream-cloud/master (because cyverse/master tries to set up LDAP which failed in this environment)
- Had to downgrade Atmosphere rtwo from 0.4.16 to 0.4.10 (see [ATMO-1644](https://pods.iplantcollaborative.org/jira/browse/ATMO-1644))
- Had to disable the 41_gateone.yml playbook (because my test environment has no GateOne server)

Under these conditions, I was able to test the following:

- [x] Log into Troposphere UI using mock auth (user "lenards")
- [x] Launch an instance (base Ubuntu 14.04 cloud image), confirm that instance launches, atmosphere-ansible runs, instance eventually enters "green light Active" state
- [x] SSH to instance with user SSH keys injected by atmosphere-ansible
- [x] Suspend instance
- [ ] Resume instance - _did not work_ - ""Timeout when waiting for search string OpenSSH in vm49.jetstreamtestcloud.cyverse.org:22" - instance resumes and shows login console in OpenStack, but does not respond to ping or SSH. I suspect this is not Atmosphere's fault.
- [x] Delete instance
- [x] Create, attach, detach, delete Volume

This was time-consuming but I felt it necessary given the amount of code I'm removing and my limited familiarity with the Atmosphere codebase. It also highlighted the need for an easy automation to create a development environment _connected to an OpenStack cloud_, in order to test a change. Another todo for me.

Questions for team leads / maintainers (perhaps @steve-gregory):
- `_get_init_script()` in service/instance.py refers to "init_files/%s/atmo-initer.rb" (which this PR removes). This function is called by `_generate_userdata_content()`, which is called by `_pre_launch_instance()`, which is called by three other functions (`boot_volume_instance()`, `launch_machine_instance()`, and `launch_volume_instance()`). Are these also no longer used?
- Is there *anything else* related to the removed files wihch I may have missed, and which should also be removed from the codebase?

julianpistorius:
An experiment I've been meaning to try:

1. Run Atmosphere with Coverage
2. Exercise it manually and/or using something like Operation Sanity
3. Check coverage report
4. Delete dead code
5. Re-run
6. Check for errors

https://coverage.readthedocs.io/

init_files/boot_scripts/hello_world.sh
c-mart: I dunno, they are literally just "hello world" in bash and Python. Is the intended audience for the "boot scripts" feature (i.e. users) likely to look in this repo?

init_files/v1/iplant_backup.sh
c-mart: According to the [Atmosphere(0) user manual](https://wiki.cyverse.org/wiki/display/atmman/Backing+Up+and+Restoring+Your+Data+to+the+Data+Store) there is an `iplant_backup` binary in the $PATH of all Atmosphere instances, and looking at a random live instance, this seems to be true. (This file is also stored in atmosphere-ansible's [atmo-idrop](https://github.com/cyverse/atmosphere-ansible/blob/c49d4fb25bf1e0881ef694608e3b49b3ae654c27/ansible/roles/atmo-idrop/files/iplant_backup.sh) role, although that doesn't appear to actually install it the script on each new instance, so I'm unsure how it gets there.) Regardless, I think that's something CyVerse-specific and doesn't belong in this repo.

c-mart:
Removed "WIP" from title, this is ready to merge from my perspective.

steve-gregory:
:shipit: Thank you @c-mart

PR #245, opened 2017-01-25
Problem: `test_a_sanity_check` still runs even when it doesn't have a…

julianpistorius:
…n appropriate test database

Solution: Changed the condition to skip when the database does _NOT_ contain the test user.

### DIFF
--- a/api/tests/v2/test_reporting.py
+++ b/api/tests/v2/test_reporting.py
@@ -1,5 +1,5 @@
 import json
-from unittest import skip, skipIf
+from unittest import skip, skipUnless
 
 from django.core.urlresolvers import reverse
 from rest_framework.test import APITestCase, APIRequestFactory, force_authenticate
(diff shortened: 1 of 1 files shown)


api/tests/v2/test_reporting.py
```diff
@@ -27,9 +27,11 @@ def setUp(self):
         self.user = UserFactory.create()
         self.view = ReportingViewSet.as_view({'get': 'list'})
 
-    @skipIf(contains_user('test-julianp'), 'The database does not contain the user test-julianp')
+    @skipIf(not contains_user('test-julianp'), 'The database does not contain the user test-julianp')
```
steve-gregory: Could also use [skipUnless](https://docs.python.org/3/library/unittest.html#unittest.skipUnless):
`@unittest.skipUnless(contains_user('test-julianp'), ...)`

julianpistorius: @steve-gregory: Done. Thanks for the suggestion.

PR #248, opened 2017-01-27
Update reporting API tests

steve-gregory:
- Include Excel renderer in the list of allowed TEST renderers.
- proof of concept worked with live data and saved xlsx file as expected


steve-gregory:
Self-:shipit: #Shameless

PR #252, opened 2017-01-31
add extra condition for resue free floating ip

xuhang57:
Currently, it doesn't know whether the floating ip has been using by another instance or not. Not necessarily to add this specific condition. 

Thanks
Lucas


xuhang57:
Sure!

PR #254, opened 2017-02-01
Problem: new versions of image fall through logic-cracks

steve-gregory:
The situation occurs when new glance images are found that represent **the second version** of an application that has been replicated across providers via an external process. These images were an unexpected edge case because the image ID is *not* the same image that created the *application UUID*, and so the logic (Testing image ID against application UUID) that blocks externally-created-images from being incorrectly ingested as atmosphere was *also* blocking these version from correctly setting the application kwargs, potentially allowing this version to create a new image. 

Instead, if REPLICATION_PROVIDER_LOCATION is set (redefined 'set' as getattr, not hasattr) attempt to find a similar image that already exists in the stack, and use its application to define yourself.


steve-gregory:
Verified and ready for merge and trickle-down to master.

PR #255, opened 2017-02-01
v2 Image API improvements

steve-gregory:
- Include the ability to include `?featured=true` to return only featured applications.
- Include default ordering to drop end-dated results to the bottom of the page, and then sort chronologically (with newest images showing up first).


julianpistorius:
👍

PR #257, opened 2017-02-02
Problem: Reporting api saves all files as reporting.xlsx

steve-gregory:
Solution:
- Allow `filename` as a query_param on the request
- Respect `filename` by saving headers to the DRF Response, *NOT*
  `StreamingHTTPResponse`


steve-gregory:
Self-:shipit:

PR #258, opened 2017-02-02
Include subnet pool specification with a default of None

steve-gregory:
**jmlowe commented on Sep 9, 2016**
> Starting with the kilo release neutron can use a shared subnet pool,
this ensures unique subnets without the need to probe. When a subnet
pool id is configured skip the search and CIDR specification and just
create the subnet using the subnet pool.
> This may need some further polish, I took the per provider override for settings at face value. One possible improvement would be to use subnet pool name and lookup the id.

**Required before merge:**
- [x] Fixed merge conflicts with master
- [x] Testing on jetstream test server using `External Network` and `External Router` topologies
- [x] Testing when `settings.DEFAULT_SUBNET_POOL_ID` is undefined, defined incorrectly, and valid.
- [x] Testing when SUBNET_POOL is preferred, but subnet was already created w/o subnetpool_id (In this case, just keep using the existing subpool)!
- [ ] Testing what happens when duplicate *critical resources* appear and how those error messages are bubbled up.

### DIFF
--- a/service/accounts/openstack_manager.py
+++ b/service/accounts/openstack_manager.py
@@ -633,11 +633,12 @@ class AccountDriver(BaseAccountDriver):
         return network_strategy
 
     def dns_nameservers_for(self, identity):
-        dns_nameservers = [
+        dns_nameservers = core_identity.provider.get_config('network', 'dns_nameservers', [])
+        db_dns_nameservers = [
             dns_server.ip_address for dns_server
             in identity.provider.dns_server_ips.order_by('order')
         ]
-        return dns_nameservers
+        return [set(db_dns_nameservers + dns_nameservers)]
 
     def delete_user_network(self, identity, options={}):
         """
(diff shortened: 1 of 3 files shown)


steve-gregory:
This branch ready for review

cdosborn:
Only quip is that names are not unique. If you search by name, you can get back multiple pools (which the code doesn't handle). The provider config should have:
```
...
subnet_pool: {
    id: ...    
    name: ...
}
...
```

jmlowe:
good call on the config by id, not accounting for name collision is a major oversight

steve-gregory:
I had aimed for `subnet pool name` because we were defining other "critical network resources" (ex: `network_name`, `router_name`) -- But i can go ahead and raise a conflict/exception if there is >1 result! 😄

steve-gregory:
- [x] Rebased to master
- [x] Remove references to `subnet_pool_name` in favor of `subnet_pool_id`

steve-gregory:
Self-:shipit: The default behavior will not be modified.

PR #261, opened 2017-02-06
Problem: /v2/instance reports incorrect 'status' and 'activity'

steve-gregory:
In the v2 API we cannot rely on 'esh_status' and 'esh_activity' because the cloud resources are not made available. As a result, some of the 'hand-created' status types like `networking, deploying, deploy_error` show up, initially incorrectly and then are 'fixed' when a call comes from the `v1` apis.

(Temporary) Solution: Return status and activity appropriately based on prior knowledge of the cloud

A longer-term solution has been suggested via FIXME: in the PR, to be discussed and then written up in a JIRA as part of the review.

Before merge Checklist:
- [x] Address the instances test case (verify the bug via test, then 'fix it'!)
- [ ] Talk to other developers about the suggested fixme and write up a new JIRA to ensure progress is not lost.
- [ ] Request review from >1 atmosphere developer.


core/models/instance_history.py
```diff
@@ -16,6 +16,20 @@ class InstanceStatus(models.Model):
...
+          suspended
+          shutoff
+          error
+          deleted
+          unknown
+    Example of 'is_final_state = False' status for Openstack:
+        - networking
+          deploying
+          deploy_error
+
```
lenards: Would having a "pending" state to show a resource in transition (i.e. - waiting on an async action within the task queue) be something to consider? Or, would that be a "faux" state for AtmoAPI use only and this is only OpenStack states?

cdosborn:
Can you explain why this is an issue:
```
As a result, some of the 'hand-created' status types like networking, deploying, deploy_error show up, initially incorrectly and then are 'fixed' when a call comes from the v1 apis.
```

lenards:
If I understand correctly ... 
I think of it is that `networking, deploy_error, deploying` as values for _activity_ and not full blown Status values. A quirk in the handling from Troposphere is that "Status = Networking" will appear with a red error dot, because any undefined status is considered to be a status that user interface was not expecting - and is, thereafter, shown as an error.

steve-gregory:
Yeah, we're sort of pulling at two loose ends of thread here, so to (hopefully) aid in understanding:

1) There was a concept in the v1 APIs of "status" vs "activity" that helped guide the "instance state workflow" through the GUI:
`active - initializing` -> `active - networking` -> `active - deploying` -> `active`

2) There is an InstanceStatusHistory class in atmosphere that registers the "Last known state" of the instance. At one point, we would route `networking` and `deploying` to a specific state (build, or active) but it was later determined that they should have their own states in order to track things like *average build time*, *average networking time*, and *average time in deployment* as separate entities.

3) The way v1 returns "Status" and "activity" relies on the cloud. the v2 APIs naively used the same methods, but without cloud-support, which resulted in unexpected results coming to the client.

cdosborn:
We could track networking time, by 'active - initializing' to 'active - newtorking'. In this case you're measuring the duration of an activity, which would be slightly different than measuring the duration of a state,  but still meaningful.

cdosborn:
So if there is no cloud object i.e. esh, we already lookup something in the history. My question is why we're hiding what's in the history. Let's fix the history if it contains bad values right?

steve-gregory:
The goal of this PR is to fix the immediate problem (that there is a UI error as a result of a mismatch of expected `status/activity` pairing in `/v1` and `/v2` APIs.

I agree with you that **something needs to be fixed**, but it would be better served to do this in another PR that is not targeted to `voracious-velociraptor`

steve-gregory:
Its true, there are lots of ways we can continue to track networking/deploying by using `activity`, but I think it creates confusion in that *all other states* are being tracked by a different attribute.

I would argue that, in a future PR, we more closely bind `status` and `activity` because one does not give the whole picture without the addition of the other.

PR #269, opened 2017-02-16
Add initial glossary and readme

cdosborn:
## Description

This is a start on a glossary for atmosphere models. I never got around to finishing it, but thought I would add the document so others could contribute.


c-mart:
@cdosborn see https://github.com/cyverse/atmosphere-guides/pull/12. There's a lot of really helpful but duplicate information between these PRs. Just need to coalesce it all in one place (and I think the right place is atmosphere-guides).

steve-gregory:
Merging into the code-base.. We can make another PR to remove this information (and point to atmosphere-guides) when `atmosphere-guides` gets these updates.

PR #277, opened 2017-02-22
Refactor: security group with neutron and nova

xuhang57:
## Description

Re-work for security group creation and rules assignment. This is the first draft. Summited for Steve @steve-gregory to review. 
 
1. might not want to use the neutron api directory in the `service/instance.py` If that's the case, could put the new functions in the `openstack_network` in rtwo.
2. might be a better way to use the json object. 
3. put all new rules in a different security group rather than `default`
4. tested with using `cloud-config` and haven't find the proper way of adding rules. 

Please let me know if I could do anything to make it more clear. 

Also, 

In order to test it, you need to change a line in `libcloud` as well. 

In  `/opt/env/atmo/lib/python2.7/site-packages/libcloud/compute/drivers/openstack.py`

```
1347         if 'ex_security_groups' in kwargs:
1348             server_params['security_groups'] = []
1349             for security_group in kwargs['ex_security_groups']:
1350                 server_params['security_groups'].append({'name': security_group})
1351                 #name = security_group.name
1352                 #server_params['security_groups'].append({'name': name})
```

Thanks,
Lucas


service/instance.py
```diff
@@ -1345,16 +1346,76 @@ def security_group_init(core_identity, max_attempts=3):
...
+        "port_range_min": None,
+        "ethertype": "IPv4",
+        "port_range_max": None,
+        "protocol": None,
+        "remote_group_id": security_group_id,
+        "security_group_id": security_group_id
+        }
+    }
+    try:
+    user_neutron.create_security_group_rule(body=first_rule)
```
julianpistorius: Hey @xuhang57 - Looks like code climate found a missing indent here.

xuhang57: Thanks Julian. (I thought I replied lol)

xuhang57:
Thanks a lot Steve. Will test it on Thursday and once I make sure everything works fine, I will merge it to my PR. 

Appreciate your help!
Lucas

service/instance.py
```diff
@@ -1345,17 +1346,98 @@ def security_group_init(core_identity, max_attempts=3):
...
 
 
-def user_security_group_init(core_identity, security_group_name='default'):
-    # Rules can come from the provider _or_ from settings _otherwise_ empty-list
-    rules = core_identity.provider.get_config('network', 'default_security_rules',getattr(settings,'DEFAULT_RULES',[]))
-    driver = get_cached_driver(identity=core_identity)
-    lc_driver = driver._connection
-    security_group = get_or_create_security_group(lc_driver, security_group_name)
-    set_security_group_rules(lc_driver, security_group, rules)
+def user_security_group_init(core_identity, security_group_name='atmosphere'):
```
xuhang57: Just one thing I want to point out is that this function was using `default` security group. I change it to use a new security group called atmosphere and that's why I add `return {"ex_metadata": ex_metadata, "ex_keyname": ex_keyname, "ex_security_groups": ex_security_groups}`

xuhang57:
Thanks for the review! @steve-gregory

service/instance.py
```diff
@@ -1608,7 +1699,8 @@ def _extra_openstack_args(core_identity, ex_metadata={}):
         # FIXME: In a new PR, allow user to select the keypair for launching
         user_key = user_keys[0]
         ex_keyname = user_key.name
-    return {"ex_metadata": ex_metadata, "ex_keyname": ex_keyname}
+    ex_security_groups = ["atmosphere"]
```
steve-gregory: We need a better way to tie this in than to do a hard-coded value.. how about something like:
```
security_group_name = core_identity.provider.get_config("network", "security_group_name", "atmosphere")
ex_security_groups = [security_group_name]
```
Similarly,[ in the area where we call user_security_group_init
](https://github.com/cyverse/atmosphere/pull/277/files#diff-545eec308259c382ab4919797a49db52R1342) we will want something similar:
```
def security_group_init(core_identity, max_attempts=3):
    has_secret = core_identity.get_credential('secret') is not None
+  security_group_name = core_identityprovider.get_config("network", "security_group_name", "atmosphere")
    if has_secret:
        return admin_security_group_init(core_identity)
    return user_security_group_init(core_identity, security_group_name=security_group_name)
```

service/instance.py
steve-gregory: Thanks for pointing that out, I missed that on initial review.

See my comment below for how we can address this 😄

service/instance.py
xuhang57: Thanks for the help and review 👍  Done.

service/instance.py
```diff
@@ -1340,22 +1341,104 @@ def user_delete_security_group(core_identity):
...
+                "port_range_min": sg_rule.get('port_range_min', None),
+                "ethertype": sg_rule.get("ethertype", "IPv4"),
+                "port_range_max": sg_rule.get('port_range_max', None),
+                "protocol": sg_rule.get("protocol", None),
+                "remote_group_id": security_group_id,
+                "security_group_id": security_group_id
+                 }
+            }
+            if 'remote_ip_prefix' in sg_rule:
+                rule_body['security_group_rule']['remote_ip_prefix'] = sg_rule['remote_ip_prefix']
```
xuhang57: ah, sorry I forgot to mention here. If `remote_ip_prefix` exists, then we need to take out the `remote_group_id` since they cannot be together in the same dictionary. I am working on it now to update the PR. @steve-gregory

steve-gregory: Thanks for pointing this out, i was not aware of that rule!

xuhang57: yeah, learn that the hard way 💯

service/instance.py
```diff
@@ -1340,22 +1341,105 @@ def user_delete_security_group(core_identity):
...
+                rule_body['security_group_rule'].pop("remote_group_id")
```
xuhang57: please let me know if you prefer a `del` .

service/instance.py
```diff
@@ -1340,22 +1341,105 @@ def user_delete_security_group(core_identity):
...
-    rules = core_identity.provider.get_config('network', 'default_security_rules',getattr(settings,'DEFAULT_RULES',[]))
-    driver = get_cached_driver(identity=core_identity)
-    lc_driver = driver._connection
-    security_group = get_or_create_security_group(lc_driver, security_group_name)
-    set_security_group_rules(lc_driver, security_group, rules)
+def user_security_group_init(core_identity, security_group_name):
+    network_driver = _to_network_driver(core_identity)
+    user_neutron = network_driver.neutron
+    extended_default_rules = _get_default_rules()
+    user_security_group_rules = core_identity.provider.get_config('network', 'user_security_rules', extended_default_rules)
```
xuhang57: I am curious, how `user_security_rules` and even the `security_group_name` are added into the `Cloud-config`? Does that need some extra work on the provider model file? @steve-gregory

xuhang57:
@steve-gregory  Thanks, Steve. 

I forgot to test without changing that line back. I wast talking about L1350-1351 in `/opt/env/atmo/lib/python2.7/site-packages/libcloud/compute/drivers/openstack.py` 

I am testing it now. I believe with using `Cloud-config`, there won't be a problem with using the original libcloud code and I totally agree with you on that.

xuhang57:
okay, so I uncommented my change in libcloud and this PR doesn't work due to that. Since even though we have security_group created by `neutron`, we still need to use liblcoud to have the security_group assigned to the instance. In other words, it is expecting a libcloud security group class but we only have a neutron object (or even None) here.

xuhang57:
@steve-gregory Added a new task called `add_security_group_task` to celery so that we can use `nova` to assign a new security group to the instance. Please let me know how I could improve this and I'd happy to refactor it. 

Thanks

xuhang57:
Will move the `add_security_group` tasks up to right after the `wait-to`. Expected to be done by Tuesday. (Currently, my laptop is sent for repair so I have to start refactoring after I have it back).

Thanks.
Lucas

xuhang57:
Blocked on writing the unit test for this. Hopefully I can get this done soon.

service/tasks/driver.py
```diff
@@ -623,6 +623,8 @@ def get_chain_from_build(
     """
     wait_active_task = wait_for_instance.s(
         instance.id, driverCls, provider, identity, "active")
+    add_security_group = add_security_group_task.si(driverCls, provider, core_identity, instance.id)
+    wait_active_task.link(add_security_group)
```
xuhang57: @steve-gregory  i add the task here.

xuhang57:
fixed conflicts

xuhang57:
@steve-gregory  no longer a WIP unless you have any suggestions. Thanks!

service/instance.py
steve-gregory: For now, you would make this change in the python REPL and/or the `/admin` section for the specific provider in question.

_In a BYOC future,_ this will be handled in a cleaner admin section within troposphere that has well-defined segments and keys.

service/instance.py
steve-gregory: `pop()` is fine and has been used elsewhere in the code to get rid of extraneous credentials/kwargs

service/tasks/driver.py
steve-gregory: Two things: One, this task is not 'chained', see the modification below. The more important 'thing' is that this task will not work when `has_secret`[1] is False. we need to do one of two things:
1 - if `core_identity` `has_secret` (see https://github.com/cyverse/atmosphere/pull/277/files#diff-545eec308259c382ab4919797a49db52R1344) then:
```
wait_active_task.link(add_security_group)
add_security_group.link(network_start)
start_chain = wait_active_task
return start_chain
```
If has_secret is False:
```
wait_active_task = ...
start_chain = wait_active_task
start_chain.link(network_start)
return start_chain
```

Also, since we're still calling `security_group_init` on instance launch, is this something we could do _before_ launching the instance, rather than after? just checking..

Where security_group_init gets called: https://github.com/xuhang57/atmosphere/blob/1174c918eb159eb116b8d5d28f8d4091b7df571a/service/instance.py#L1711

[1] - We really need a better feature-flag for providers that "have admin accounts" or "do not have admin account and are run by the user".

service/instance.py
xuhang57: great! Good to know that. Thanks!

service/instance.py
xuhang57: agree. I just hope it would not consume too much memory

service/tasks/driver.py
xuhang57: Good find! Thanks! I have updated a new patch for this.

And for the `security_group_init`, we still need to have it since we need to create a security group for the user during the first launch. And that's something we could do before launching the instance. And assigning security group, which is handled by celery can only do after launching the instance. HTH. 👍

xuhang57:
@steve-gregory  Thanks for the review. Please see comments 👍

xuhang57:
Hold on for now. I am looking into some `rtwo` issues. Seems to have some problems with creating a UserManager

xuhang57:
Depends on : https://github.com/cyverse/rtwo/pull/11

xuhang57:
@steve-gregory  Resolved :) Not sure if this could be in the Y-release?

service/instance.py
```diff
@@ -1346,22 +1347,106 @@ def user_delete_security_group(core_identity):
 
 def security_group_init(core_identity, max_attempts=3):
     has_secret = core_identity.get_credential('secret') is not None
+    security_group_name = core_identity.provider.get_config("network", "security_group_name", "atmosphere")
```
steve-gregory: One minor change, lets swap `atmosphere` for `default`

service/instance.py
```diff
@@ -1601,6 +1729,7 @@ def _extra_openstack_args(core_identity, ex_metadata={}):
         # FIXME: In a new PR, allow user to select the keypair for launching
         user_key = user_keys[0]
         ex_keyname = user_key.name
+    security_group_name = core_identity.provider.get_config("network", "security_group_name", "atmosphere")
```
steve-gregory: `atmosphere`->`default`

service/tasks/driver.py
```diff
@@ -1219,6 +1225,30 @@ def check_web_desktop_task(driverCls, provider, identity,
...
+    """
+    from service.instance import _to_user_driver
+    user_driver = _to_user_driver(core_identity)
+    user_nova = user_driver.nova
+    try:
+        server_instance = user_nova.servers.get(instance_alias)
+    except:
+        raise Exception("Cannot find the instance")
+    try:
+        security_group_name = core_identity.provider.get_config("network", "security_group_name", "atmosphere")
```
steve-gregory: `atmosphere`->`default`

xuhang57:
@steve-gregory changed! Thanks. Let's see whether the CI works

steve-gregory:
:shipit:

PR #279, opened 2017-02-24
Problem: using deprecated method for creating keystoneclient

steve-gregory:
## Solution

Fix rtwo, chromogenic, and the args that are used to instantiate the two to provide the expected parameters for keystoneauth.

This new method should avoid "Bad service catalogs" when creating drivers as the admin/user, and avoid `404 Endpoint not found` errors that were being found in the current release of Atmosphere.

- [x] Tested against the production cloud(s) including `Cyverse Cloud`
- [x] Tested against the jetstream cloud(s)


steve-gregory:
Tested and verified on atmobeta.

PR #284, opened 2017-03-09
WIP: application_to_provider.py

c-mart:
## Description
This script makes an Application (a.k.a. image) available on a specified new
provider by doing any/all of the following as needed:

 - Populates Glance image metadata
 - Transfers image data from existing provider
   - Image data migrated using Glance API or iRODS for Atmosphere(0)
 - (Future coming soon) if Application uses an AMI-style image, ensures the
   kernel (AKI) and ramdisk (ARI) images are also present on destination
   provider, and sets appropriate properties
 - Creates models (ProviderMachine, InstanceSource) in Atmosphere database

Gracefully handles the case where destination provider is already partially
populated with image data/metadata (only missing information will be added).

## Background
This script is required for Atmosphere to support a common application across multiple cloud providers that run heterogenous versions of OpenStack, i.e. when we cannot do a periodic "background Glance database sync". Specifically, this script enables us to make our Applications available on the new Marana Cloud for Atmosphere(0).

## Checklist before merging
- [ ] Script should handle AMI-style images with dependent AKI (kernel) and ARI (ramdisk) images
- [ ] Address todos in code
- [ ] Address remaining questions in [blueprint](https://github.com/cyverse/atmosphere/wiki/Blueprint-for-Atmosphere(2)-script,-%22application_to_provider.py%22)
- [ ] Perhaps support a "background iRODS" mode of image data transfer, for Atmosphere(0)
- [ ] New test(s) included to verify the script works as intended
- [ ] Documentation created/updated as needed
- [ ] Reviewed and approved by at least one other contributor


c-mart:
After checking with @steve-gregory, proposing this as it is for W-release. Support for using iRODS for data transfer to be added in a later iteration.

PR #299, opened 2017-03-22
Small updates from the Jetstream server

steve-gregory:
## Description
These unstaged changes were on the jetstream server and should be included before the next release.

**Highlights:**
- Bugfix: Include domain_name for non-2.0 password auth (on openstack clouds)
- Internal: Increase time limit and duration for allocation monitoring tasks.
- Bugfix: Updated 'share_image' to pass Identity (for all credentials) rather than just project_name.

### DIFF
--- a/core/models/machine_request.py
+++ b/core/models/machine_request.py
@@ -297,6 +297,8 @@ class MachineRequest(BaseRequest):
         old_admin = old_provider.get_admin_identity().get_credentials()
         if 'ex_force_auth_version' not in old_creds:
             old_creds['ex_force_auth_version'] = '2.0_password'
+        if old_creds['ex_force_auth_version'] != '2.0_password' and 'domain_name' not in old_creds:
+            old_creds['domain_name'] = 'default'
         old_creds.update(old_admin)
 
         new_provider = self.new_machine_provider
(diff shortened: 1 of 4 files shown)


core/models/machine_request.py
```diff
@@ -297,6 +297,8 @@ def get_credentials(self):
         old_admin = old_provider.get_admin_identity().get_credentials()
         if 'ex_force_auth_version' not in old_creds:
             old_creds['ex_force_auth_version'] = '2.0_password'
+        if old_creds['ex_force_auth_version'] != '2.0_password' and 'domain_name' not in old_creds:
```
julianpistorius: Note: I had to read through this method a couple of times. The name (`get_credentials`) is a bit deceptive. It does quite a bit more than just 'getting'. It seems to do some clever copying and manipulation. I imagine it started out simpler.

service/accounts/openstack_manager.py
```diff
@@ -441,30 +441,33 @@ def shared_images_for(self, image_id):
                     for member in shared_with]
         return projects
 
-    def share_image_with_project(self, glance_image, project_name):
```
julianpistorius: Not part of this PR, but I noticed that lines 430 & 431 contain tabs for indentation.

julianpistorius: https://github.com/steve-gregory/atmosphere/pull/6

steve-gregory: :shipit:

PR #300, opened 2017-03-27
Remove ScriptDeployment from mount_volume tasks

steve-gregory:
## Description

- [x] Support for ext4 as default filesystem type.
- [x] Allow overriding the default filesystem type via cloud config option: `provider.cloud_config["deploy"]["volume_fs_type"]`
- [x] Allow overriding the default volume mount prefix `/vol_` via cloud config option: `provider.cloud_config['deploy']['volume_mount_prefix']`
- [x] Ansible modules over ScriptDeployment for 'checking' and 'mounting' the volume.

## Related PRs
https://github.com/cyverse/atmosphere-ansible/pull/79

### DIFF
--- a/service/deploy.py
+++ b/service/deploy.py
@@ -76,6 +76,24 @@ def ready_to_deploy(instance_ip, username, instance_id):
         limit_playbooks=['check_networking.yml'])
 
 
+def deploy_check_volume(instance_ip, username, instance_id,
+        device, device_type='ext4'):
+    """
+    Use service.ansible to deploy to an instance.
+    """
+    extra_vars = {
+        "VOLUME_DEVICE": device,
+        "VOLUME_DEVICE_TYPE": device_type,
+    }
+    playbooks_dir = settings.ANSIBLE_PLAYBOOKS_DIR
+    playbooks_dir = os.path.join(playbooks_dir, 'utils')
+    limit_playbooks = ['check_volume.yml']
+    return ansible_deployment(
+        instance_ip, username, instance_id, playbooks_dir,
+        limit_playbooks=limit_playbooks,
+        extra_vars=extra_vars)
+
+
 def instance_deploy(instance_ip, username, instance_id,
 		    limit_playbooks=[]):
     """
--- a/service/deploy.py
+++ b/service/deploy.py
@@ -98,7 +98,7 @@ def deploy_mount_volume(instance_ip, username, instance_id,
 def deploy_check_volume(instance_ip, username, instance_id,
         device, device_type='ext4'):
     """
-    Use service.ansible to deploy to an instance.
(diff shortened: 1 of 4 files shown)


service/deploy.py
```diff
@@ -76,6 +76,43 @@ def ready_to_deploy(instance_ip, username, instance_id):
...
+    return ansible_deployment(
+        instance_ip, username, instance_id, playbooks_dir,
+        limit_playbooks=limit_playbooks,
+        extra_vars=extra_vars)
+
+
+def deploy_check_volume(instance_ip, username, instance_id,
+        device, device_type='ext4'):
+    """
+    Use service.ansible to deploy to an instance.
```
c-mart: I'm not sure what you mean by "deploy to an instance".. this function appears to just run check_volume.yml against the instance?

service/tasks/volume.py
```diff
@@ -30,45 +31,33 @@
...
+        # One playbook to make two checks:
+        # 1. Voume exists
+        # 2. Volume has a filesystem
+        #    (If not, create one of type 'device_type')
+        playbooks = deploy_check_volume(
+            instance.ip, username, instance.id,
+            device_location, device_type=device_type)
+        celery_logger.info(playbooks.__dict__)
+        hostname = build_host_name(instance.id, instance.ip)
+        result = False if execution_has_failures(playbooks, hostname)\
```
c-mart: This is confusing to me but perhaps because of my ignorance. Instead of setting `result = False` and raising a generic exception, why wouldn't we raise exceptions close to where they happen (e.g. inside of whatever `execution_has_failures()` does) and catch / re-raise them here, maybe exposing more error detail to the API client?

steve-gregory: Because the task is happening asynchronously, we no longer have the opportunity to communicate with the API...

Traditionally when we run playbooks, we need to know the success/failure before moving on to the next task and this was seen as a relatively easy way of verifying that before moving forward.

I'd think that updating the behavior of `execution_has_failures` to something similar to what you described above would be a fine idea for another PR/JIRA ticket

c-mart: Got it. Thanks!

c-mart:
@steve-gregory I am concerned that with this PR, we're now using Ansible to mount volumes but not un-mount them. This means that Ansible will persist the mount to `/etc/fstab`, but the persistent mount will not be removed from `/etc/fstab` when the volume is detached. I fear that after a volume is detached, if the instance is rebooted then it will not come back because of failure to mount a nonexistent volume. Adding a test case to the deployment plan tomorrow, there we'll find out for sure.

c-mart:
Also, is [this](https://github.com/cyverse/atmosphere/blob/master/service/deploy.py#L417) now dead code that should be removed?

PR #301, opened 2017-03-28
Problem: Travis.ci is not building for Atmosphere

steve-gregory:
## Solution:
 Include travis.ci in atmosphere by providing a .travis.yml

## Highlight Reel:
- [x] Include redis-server and postgres 9.6 in travis.yml configuration
- [x] Updated variables.ini.dist to use '#' for comments instead of ';'
- [x] Provide default variables for non-string values.. Although configure doesn't respect them..
- [x] Remove unused variables from .j2 and variables.ini.dist
- [x] Update configure: Strip all comments (#) from the line to be evaluated


steve-gregory:
Self-:shipit: -- Let there be travis!

PR #309, opened 2017-03-30
Problem: Duplicate allocation sources

julianpistorius:
# Solution: 
- Use `allocation_source_name` as the unique key
- Include events that mark when renewal occurred
- More events for keeping track of user allocation sources

## Note

Needs to go along with the Troposphere counterpart: https://github.com/cdosborn/troposphere/tree/ww-jetstream-fixes


api/v2/serializers/details/allocation_source.py
```diff
@@ -63,6 +63,6 @@ def get_updated(self, allocation_source):
     class Meta:
         model = AllocationSource
         fields = (
-            'id', 'url', 'name', 'source_id', 'compute_allowed',
```
steve-gregory: Is a change in troposphere required to reflect the change in name here?

core/migrations/0075_remove_allocationsource_source_id.py
```diff
@@ -0,0 +1,19 @@
...
+
+    dependencies = [
+        ('core', '0074_unique_allocationsource_uuid'),
+    ]
+
+    operations = [
+        migrations.RemoveField(
+            model_name='allocationsource',
+            name='source_id',
+        ),
```
steve-gregory: Should be compressed into `0074_.py` unless an error occurs.

core/migrations/0076_unique_allocationsource_names_part1.py
```diff
@@ -0,0 +1,44 @@
...
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ('core', '0075_remove_allocationsource_source_id'),
+    ]
+
+    operations = [
+        migrations.RunPython(
+            remove_duplicate_allocation_sources,
+            reverse_code=migrations.RunPython.noop)
```
steve-gregory: Should be compressed into `0074_.py` unless an error occurs.

jetstream/allocation.py
```diff
@@ -255,32 +257,48 @@ def get_user_allocations(self, username, include_expired=False, raise_exception=
...
-        return source, False
-    except AllocationSource.DoesNotExist:
-        source = AllocationSource.objects.create(
-            name=source_name,
-            compute_allowed=compute_allowed,
-            source_id=source_id
-        )
-        return source, True
+        created_event_key = 'sn=%s,si=%s,ev=%s,dc=jetstream,dc=atmosphere' % (
+            source_name, source_id, 'allocation_source_created_or_renewed')
```
steve-gregory: Looks like this format is used in a bunch of places in the file -- maybe we should make a method for this?

jetstream/tasks.py
```diff
@@ -136,69 +132,58 @@ def send_reports():
...
+
+            created_or_updated_event = EventTable.objects.filter(name='allocation_source_created_or_renewed',
+                                                                        payload__allocation_source_name=allocation_source.name).order_by('timestamp').last()
+
+            if created_or_updated_event:
+                #if renewed, change ignore old allocation usage
+                start_date = created_or_updated_event.payload['start_date']
+
+            for user in allocation_source.all_users:
+                compute_used, burn_rate = total_usage(user.username, start_date, allocation_source_name=allocation_source.name ,end_date=end_date,burn_rate=True)
```
steve-gregory: ` ,` 😉

Seriously though -- Looks much cleaner than the previous iteration 👍

core/migrations/0076_unique_allocationsource_names_part1.py
```diff
@@ -0,0 +1,44 @@
...
+            cursor.execute('DELETE FROM allocation_source_snapshot WHERE allocation_source_id = %s',
+                           [allocation_source_id])
+            cursor.execute('DELETE FROM instance_allocation_source_snapshot WHERE allocation_source_id = %s',
+                           [allocation_source_id])
+            cursor.execute('DELETE FROM user_allocation_snapshot WHERE allocation_source_id = %s',
+                           [allocation_source_id])
+            cursor.execute('DELETE FROM user_allocation_source WHERE allocation_source_id = %s',
+                           [allocation_source_id])
+            cursor.execute('DELETE FROM allocation_source WHERE id = %s', [allocation_source_id])
+
```
lenards: 📝 To summarize the purpose here, if we didn't remove those records from these tables - then the next migration would *not* succeed as there _could_ be records that violate the constraint added.

core/migrations/0077_unique_allocationsource_names_part2.py
```diff
@@ -0,0 +1,19 @@
...
+        ('core', '0076_unique_allocationsource_names_part1'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='allocationsource',
+            name='name',
+            field=models.CharField(editable=False, max_length=255, unique=True),
+        ),
+    ]
```
lenards: Now that all those potential duplicates are gone (see migration 76), we can add the constraint.

core/models/event_table.py
```diff
@@ -52,10 +54,14 @@ def listen_for_changes(sender, instance, created, **kwargs):
     return None
 
 # Instantiate the hooks:
-pre_save.connect(listen_before_allocation_snapshot_changes, sender=EventTable)
-post_save.connect(listen_for_allocation_overage, sender=EventTable)
+#pre_save.connect(listen_before_allocation_snapshot_changes, sender=EventTable)
+#post_save.connect(listen_for_allocation_overage, sender=EventTable)
 # post_save.connect(listen_for_changes, sender=EventTable)
+#post_save.connect(listen_for_allocation_snapshot_changes, sender=EventTable)
+#post_save.connect(listen_for_user_snapshot_changes, sender=EventTable)
```
lenards: Do we want to straight-out remove this hooks? 
I would vote "yes" 👍

lenards: It would be good as an X-X thing - no need now.

jetstream/allocation.py
```diff
@@ -338,19 +354,58 @@ def fill_user_allocation_sources():
...
-            user=user)
+        allocation_source = get_or_create_allocation_source(api_allocation)
         allocation_resources.append(allocation_source)
+        user_allocation_source = get_or_create_user_allocation_source(user, allocation_source)
+        user_allocation_sources.append(user_allocation_source)
+
+    canonical_source_names = [source.name for source in allocation_resources]
+    for user_allocation_source in old_user_allocation_sources:
+        if user_allocation_source.allocation_source.name not in canonical_source_names:
+            delete_user_allocation_source(user, user_allocation_source.allocation_source)
```
lenards: This is beyond a "Tuesday" thing, but is it okay if a `user_allocation_source` within `old_user_allocation_sources` causes a failure that would _prevent_ other "user_allocation_source"s from being processed? 

If I have a list, like, say [asrc11, asrc12, asrc13, asrc14, asrc15, asrc16, asrc17] - and asrc12 caused a failure, then we don't know if asrc13-17 should be deleted? 

Maybe my question is gibberish.

scripts/add_renewal_events.py
```diff
@@ -0,0 +1,46 @@
...
+
+# fetch values from TACC
+api = TASAPIDriver()
+allocations = api.get_all_projects()
+
+for source in allocations:
+    for details in source['allocations']:
+
+        hashstring = '%s_%s' % (source['chargeCode'], details['id'])
+        hashed_uuid = uuid.uuid5(uuid.NAMESPACE_DNS, str(hashstring))
```
lenards: Should this be `uuid.NAMESPACE_X500`? 

If so, I think the `hashstring`'s format would need to change - no?

lenards: For reference, the format string used in `jetstream/allocation.py` is:
```
'sn=%s,si=%s,ev=%s,dc=jetstream,dc=atmosphere' 
```

scripts/update_event_table.py
```diff
@@ -0,0 +1,57 @@
...
+        post_save.connect(listen_for_instance_allocation_changes, sender=EventTable)
+        post_save.connect(listen_for_allocation_source_created_or_renewed,sender=EventTable)
+        post_save.connect(listen_for_allocation_source_compute_allowed_changed,sender=EventTable)
+
+    else:
+        post_save.disconnect(listen_for_allocation_threshold_met, sender=EventTable)
+        post_save.disconnect(listen_for_instance_allocation_changes, sender=EventTable)
+        post_save.disconnect(listen_for_allocation_source_created_or_renewed, sender=EventTable)
+        post_save.disconnect(listen_for_allocation_source_compute_allowed_changed, sender=EventTable)
+
```
lenards: This is missing "signals" added as hooks in `core/models/event_table.py`:
```
 +post_save.connect(listen_for_user_allocation_source_created, sender=EventTable)
 +post_save.connect(listen_for_user_allocation_source_deleted, sender=EventTable)
```

api/v2/serializers/details/allocation_source.py
lenards: I asked a similar question, I think that [Troposphere PR 566](https://github.com/cyverse/troposphere/pull/566) addresses the change from `..source_id` to `..name`. 

The only curious thing is that Connor has a fix for an issue with `SENTRY_ENABLED` on that _fixes_ branch that we haven't pulled into W-W (see here: [f279cd407](https://github.com/cdosborn/troposphere/commit/f279cd4076a1ecb2dc40ee2c3572ebf8c8bd675e)).

jetstream/allocation.py
lenards: I'm 👍 on a method defined. 

I would consider having that method use a format string definition of this as well

jetstream/allocation.py
```diff
@@ -255,32 +257,48 @@ def get_user_allocations(self, username, include_expired=False, raise_exception=
...
+        created_event_key = 'sn=%s,si=%s,ev=%s,dc=jetstream,dc=atmosphere' % (
+            source_name, source_id, 'allocation_source_created_or_renewed')
+        created_event_uuid = uuid.uuid5(uuid.NAMESPACE_X500, str(created_event_key))
+        created_event = EventTable.objects.create(name='allocation_source_created_or_renewed',
+                                                  uuid=created_event_uuid,
+                                                  payload=payload)
+        assert isinstance(created_event, EventTable)
+    except IntegrityError as e:
+        # This is totally fine. No really. This should fail if it already exists and we should ignore it.
+        pass
```
lenards: This is calling attention to the fact that the normal operation of the consuming API might produce such errors, and that they can be ignored. 

How? It seems that the polling frequency is such that responses may appear that would create an event - or update an event - depending on the first occurrence seen.

jetstream/allocation.py
lenards: here is a slight variation in the formats, `ca=%2` for "compute allowed" and absent in the above occurrence.

jetstream/allocation.py
```diff
@@ -255,32 +257,48 @@ def get_user_allocations(self, username, include_expired=False, raise_exception=
...
+        compute_event_uuid = uuid.uuid5(uuid.NAMESPACE_X500, str(compute_event_key))
+        compute_allowed_event = EventTable.objects.create(
+            name='allocation_source_compute_allowed_changed', uuid=compute_event_uuid, payload=payload)
+        assert isinstance(compute_allowed_event, EventTable)
+    except IntegrityError as e:
+        # This is totally fine. No really. This should fail if it already exists and we should ignore it.
+        pass
+
+    source = AllocationSource.objects.get(name__iexact=source_name)
+    return source
```
lenards: One way of looking at this method is that we have implemented an "append-only" sort of idempotent operation:
- where the first occurrence will generate two events when there is no previous event
- where the second occurrence will generate one event when there has been a previous event

lenards:
> Looks good generally -- Could you comment on when/how `scripts/add_renewal_events.py` 
> and `scripts/update_event_table.py` should be used (and what order if thats important)?

@steve-gregory - I asked similar questions about these ... 

It seems like they are logic that need to be included in migrations, but we do not plan on running them more than once. 

Those facts would seem to eliminate them from being considered for _inclusion_ in `./scripts`.

scripts/add_renewal_events.py
```diff
@@ -0,0 +1,46 @@
...
+
+        payload = {
+            "start_date": details['start'],
+            "end_date": details['end'],
+            "compute_allocated": details['computeAllocated'],
+            "allocation_source_name": source['chargeCode'],
+        }
```
lenards: Julian said "this would have happened in schema validation for `payload` had gotten in"

scripts/add_renewal_events.py
```diff
@@ -0,0 +1,46 @@
...
+
+            e = EventTable(
+                uuid=hashed_uuid,
+                name="allocation_source_created_or_renewed",
+                payload=payload,
+                entity_id=payload['allocation_source_name'],
+                timestamp=payload['start_date'])
+            e.save()
+        except:
+            continue
```
amit4111989: So this script adds create_or_renewal event for all the existing allocation sources (update_snapshot in jetstream/tasks.py looks at this event for a start date before computing user_allocation_snapshots)

 if an event already exists then we will get an error because of the unique UUID constraint. This would also mean we do not need to create an event  for the allocation source at all. 
But this script is not in use anymore and can be deleted. This was converted to migration 0079 yesterday

steve-gregory:
:shipit:

PR #318, opened 2017-04-06
iRODS transfer support for application_to_provider

c-mart:
This PR adds support for image data transfer via iRODS data object copy, when both the source and destination providers both use the [iRODS storage back-end for Glance](https://github.com/cyverse/glance-irods), and both providers share a common iRODS zone. Very specific to CyVerse atmosphere(0) and perhaps not likely that anyone else will use it! The benefit for us is that migrating images becomes very fast, on the order of a minute rather than 10-30+ minutes.

Using iRODS to transfer image data requires us to reach "around" the Glance API, populating the image data in iRODS out-of-band, and then adding the location to the image object in the Glance API. Unfortunately, using Glance in this way has some caveats:
- Glance does not allow adding a location to an image unless we enable a feature (`show_multiple_locations`) which presents a security issue for us: it would expose a connection string to users which contains credentials for the iRODS service account used by Glance server. For now, commenting out [three lines of code](https://github.com/openstack/glance/blob/master/glance/api/v2/images.py#L294-L297) fixes this limitation. Such a patch will no longer be necessary in OpenStack Pike release, when [`show_multiple_locations` is deprecated](https://github.com/openstack/glance/blob/85eaa3929e5c41396d7c7139446ea9dd098da421/releasenotes/notes/deprecate-show-multiple-location-9890a1e961def2f6.yaml) and instead we can use `policy.json` to control who can set/get image locations.
- The image checksum is not populated in the Glance database when the Glance API is not used to upload image data, so we end up with Glance images that have a checksum of `None`.
  - This means that when we transfer image data using iRODS, `application_to_provider.py` can no longer use checksums to determine if image data needs to be transferred. It needs to use the image size, which is not a strong assurance that the bits match.
  - This could be changed in glance, see [my comment on bug 1551498](https://bugs.launchpad.net/glance/+bug/1551498/comments/4).

## Checklist before merging
- [x] test final changeset against a fresh development environment, ensure that images migrated using iRODS are launchable
- [x] resolve python-irodsclient dependency in rtwo
- [ ] ~ideally get [irods/python-irodsclient #67](https://github.com/irods/python-irodsclient/pull/67) merged so we don't need to use my fork~ (not waiting for this to happen)
- [ ] ~Documentation created/updated at [Example link to documentation](https://example.test/doc#new_section) to give context to the feature~ (script already referenced [here](https://github.com/cyverse/atmosphere-guides/pull/17/files) and iRODS support options are documented in script help)
- [x] If necessary, include a snippet in CHANGELOG.md

## Checklist after merging
- [ ] For a maintainer: backport change to whymsical-wyvern

### DIFF
--- a/scripts/application_to_provider.py
+++ b/scripts/application_to_provider.py
@@ -422,9 +422,6 @@ def migrate_image_data_glance(src_glance_client, dst_glance_client, img_uuid, lo
 
 
 def migrate_image_data_irods(dst_glance_client, irods_conn, irods_src_coll, irods_dst_coll, img_uuid):
-
-    # TODO test me and add exception handling! Test me in sad cases (e.g. bad creds, could not make connection)
-
     sess = iRODSSession(host=irods_conn.get('host'),
                         port=irods_conn.get('port'),
                         zone=irods_conn.get('zone'),
--- a/scripts/application_to_provider.py
+++ b/scripts/application_to_provider.py
@@ -439,6 +439,19 @@ def migrate_image_data_glance(src_glance_client, dst_glance_client, img_uuid, lo
 
 
 def migrate_image_data_irods(dst_glance_client, irods_conn, irods_src_coll, irods_dst_coll, img_uuid):
+    """
+    Migrates image data using iRODS and then sets image location using Glance API.
+
+    Args:
+        dst_glance_client: glance client object for destination provider
+        irods_conn: dict as returned by _parse_irods_conn()
+        irods_src_coll: Path to collection for iRODS images on source provider
+        irods_dst_coll: Path to collection for iRODS images on destination provider
+        img_uuid: UUID of image to be migrated
+
+    Returns:
+
+    """
     sess = iRODSSession(host=irods_conn.get('host'),
                         port=irods_conn.get('port'),
                         zone=irods_conn.get('zone'),
@@ -459,6 +472,7 @@ def migrate_image_data_irods(dst_glance_client, irods_conn, irods_src_coll, irod
     # Assumption that iRODS copy will always be correct+complete, not inspecting checksums afterward?
     dst_glance_client.images.add_location(img_uuid, dst_img_location, dict())
     logging.info("Set image location in Glance")
+    
 
 
 def _parse_irods_conn(irods_conn_str):
--- a/scripts/application_to_provider.py
+++ b/scripts/application_to_provider.py
@@ -432,7 +432,7 @@ def migrate_image_data_glance(src_glance_client, dst_glance_client, img_uuid, lo
                 logging.warning("Image data upload attempt failed")
 
     if src_img.checksum != dst_glance_client.images.get(img_uuid).checksum:
-        raise Exception("Could not upload image data")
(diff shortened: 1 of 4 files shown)


c-mart:
Tested this in a fresh local development environment. In doing so, ran into several other issues which ended up taking most of the day. Still, every application I migrated using the iRODS transfer completed successfully and quickly. The longest one (an AMI-based image with kernel and ramdisk image dependencies) took 2 minutes, the others all under a minute. The script populated what was expected, both in iRODS, Glance, and in the Atmosphere database. Instances do launch, at this point I can't SSH to them but I'm fairly sure that's an unrelated problem.

Removing WIP label, ready to review! If we are OK with the above caveats, I would like to get this backported to WW and hotfixed to production so I can start bulk-migrating images to the Marana cloud.

scripts/application_to_provider.py
```diff
@@ -255,15 +293,42 @@ def main():
 
         local_storage_dir = secrets.LOCAL_STORAGE if os.path.exists(secrets.LOCAL_STORAGE) else "/tmp"
         local_path = os.path.join(local_storage_dir, sprov_img_uuid)
+
+        def migrate_image_data(img_uuid):
+            # Todo this function is in an awkward place and relies on 'global' state, unsure of best way to refactor
```
steve-gregory: Yeah this makes `main()` pretty turse -- Let's move this function out and isolate the arguments you require.

scripts/application_to_provider.py
```diff
@@ -36,6 +40,26 @@
...
 destination provider, script will exit with error unless
 --ignore_missing_members is set.
+
+The iRODS transfer feature was developed for CyVerse Atmosphere(0); may be of
+limited use elsewhere. In order to use it:
+- Source and destination providers must use the iRODS storage backend for
+  OpenStack Glance (https://github.com/cyverse/glance-irods)
+- Src. and dst. providers must store images in the same iRODS zone
+- --source-provider-id, --irods-conn, --irods-src-coll, and --irods-dst-coll
+  must all be defined
```
steve-gregory: Would seem that only _one_ of the two collections would be required if someone wanted to go `iRODS -> Glance` or `Glance -> iRODS`?

Just curious how far we are away from Many-to-many relationship 😄

c-mart: @steve-gregory, I don't know what you mean. Downloading an image from provider A using iRODS and then uploading it to provider B using Glance, or vice versa? I haven't built support for that. Regardless of which storage back-end is in use, why would someone want to do that over the default way of using the Glance API for the entire migration? The only reason to use iRODS transfer is for the speedup of iRODS data object copy, and for that you wouldn't want to be uploading *or* downloading anything via Glance API.

I don't know what you mean by "many-to-many relationship". Can you please clarify?

scripts/application_to_provider.py
c-mart: @steve-gregory did this in commit 9440eaa.

c-mart:
Made script more functional and a few other improvements. Tested to ensure it still migrates images both using iRODS transfer and pure Glance API transfer. Feel free to merge if there are no other requested changes applicable to this PR.

PR #340, opened 2017-04-18
Problem: Allocation engines are different for Cyverse/Jetstream

steve-gregory:
## Solution
This pull request merges the way th that Jetstream tracks resource usage and allows Cyverse the ability to assign allocations and track resource usage in atmosphere. This PR will create an allocation source model which will act as the primary source of allocations for users. The allocation sources can be created and modified by an admin user, and will be shared between the users added to these sources (by the admin).

Each machine (instance) of a user will have to be assigned to an allocation source in order to launch the instance. The allocation sources will renew based on their renewal strategy which can be defined by the rules engine, which is a new addition to the atmosphere backend and will help create and regulate business rules for allocation source management.

All of the allocation source creation, modification and destruction will be managed with v2 apis in conjunction with an event table, which is essentially an event sourcing model that acts as a ledger for allocation source related actions and events.

To help validate the calculations and operations of the allocation source model, we have also introduced django-behave which is a BDD testing plugin for django that facilitates writing useful and unique test cases for each scenario in human readable language (english). There are currently 17 scenarios and 72 steps included in this PR which test every aspect of the allocation source model, and spoof user and instance interactions with the allocations for usage calculation purposes.

## Workflows that have been verified

- [x] Migration to new system: Create an individual allocation source for each user, assign each source to each user.
- [x] New user onboarding: Upon account creation, create an individual allocation source and assign to the user.
- [x] Existing user requests additional resources: Admins will change compute_allowed in the admin panel, which will update accordingly.
- [x] Existing user, monthly renewal: Taken care of by 'update_snapshot', rules engine, + 'default' AllocationSource.strategy
- [x] Support for Shared usage of Allocation sources (Limited allocation aka "Special Allocations")
- [x] Support for "threshold" emails (ToDo: Feature 🏳️  this!)
- [x] Enforcement (done periodically and preventing instance launch/action)
- [x] End-dating of allocation sources is supported (deletion of allocation source is not allowed)
- [x] Jetstream Migrations: AllocationSource events and code have changed to match jetstream.

### Skipped for this PR
- [ ] TODO: Currently, no INSTANCES are assigned to the individual's allocation. This should be done to allow an effective transfer to the new system.
- [ ] TODO: Don't allow End-date (Deletion) of allocation source -- Not allowed if currently running instances still exist. (All instances assigned to this allocation source should have end-dates!)
- [ ] Future: Instances should be suspended upon allocation source (end dating/deletion)
- [ ] Future: Instances should be able to "set a new Allocation Source" before starting the resource again.
- [ ] Future: Carry-over/Rollover with Allocation Sources

- [ ] Troposphere Future: Single-selection of AllocationSource *should* be moved automatically, rather than showing a "Selection Modal"
- [ ] Troposphere Future: Single-selection of AllocationSource *should not* be shown in the Instance Details page (Because there is nothing to select!)
- [ ] 

- [ ] All work in https://github.com/orgs/cyverse/projects/2 is completed prior to merge


cyverse_allocation/views.py
```diff
@@ -0,0 +1,3 @@
+from django.shortcuts import render
+
+# Create your views here.
```
steve-gregory: Structure comment: we may want to `cyverse_allocation` -> `allocation` and toss the existing allocation structure as part of this PR.

api/v2/serializers/details/renewal_strategy.py
```diff
@@ -0,0 +1,33 @@
...
+        return strategy[1]['id']
+
+    def get_compute_allowed(self,strategy):
+        return strategy[1]['compute_allowed']
+
+    def get_renewed_in_days(self,strategy):
+        return strategy[1]['renewed_in_days']
+
+    def get_external(self, strategy):
+        return strategy[1]['external']
```
amit4111989: This isnt the best way to do it , but I was trying to serialize values from a dictionary instead of a model, and the index 1 represents that dict while index 0 is the name of the strategy. This helped me form a defined structure for the rules dict ( used with rules engine )

api/v2/views/allocation_source_command.py
```diff
@@ -0,0 +1,263 @@
...
+            raise Exception('Renewal Strategy %s is not valid'%(renewal_strategy))
+        return True
+
+    def _for_validate_allocation_source(self,source_id):
+        # raise Exception('Error with Source ID')
+        return True
+
+    def _for_validate_name(self,name):
+        # raise Exception('Error with Allocation Source Name')
+        return True
```
amit4111989: This is dead code. I will take it out with another PR

amit4111989:
Looks good. As Steve mentioned, some helper functions for Allocation Source may be best suited in core/models/allocation_source . Another PR in the future should take care of it and some dead code I spotted

PR #346, opened 2017-04-24
Problem: 'Redeploy' InstanceAction created incorrectly in migrations

steve-gregory:
## Solution:

Include the 'redeploy' key in the migration. 

Have you already applied this migration? Go ahead and run these lines:
```
from core.models import InstanceAction
action = InstanceAction.objects.get(name='Redeploy')
action.key = 'Redeploy'
action.save()
```


steve-gregory:
@xuhang57 FYI you may need to follow the instructions listed above (You likely already have this migration in the books)

PR #357, opened 2017-05-04
Include Web Desktop signature generation inside Atmosphere

steve-gregory:
... exposed via /v2/web_tokens/<instance_id>

> where `instance_id` is an identifier of type UUID field, the `provider_alias` for the instance

## Description

[This PR depends on a Troposphere PR](https://github.com/cyverse/troposphere/pull/611).


api/v2/views/web_token.py
```diff
@@ -0,0 +1,88 @@
...
+        client_ip = request.META['REMOTE_ADDR']
+        auth_token = request.session.get('token')
+
+        logger.info("ip_address: %s" % ip_address)
+        logger.info("client_ip: %s" % client_ip)
+        client_ip_fingerprint = SIGNER.get_signature(client_ip)
+        token_fingerprint = SIGNER.get_signature(auth_token)
+
+        sig = SIGNED_SERIALIZER.dumps([ip_address,
+            client_ip_fingerprint,
```
julianpistorius: We don't need `client_ip_fingerprint`. By relying on it I caused the problem with connection problems on some wifi networks. Also it kinda allows a bad actor to directly provide data to be signed (spoofing).

api/v2/views/web_token.py
```diff
@@ -0,0 +1,88 @@
...
+            token_fingerprint])
+
+        #FIXME: stop here? Remove fingerprinting?
```
julianpistorius: Remove comment.

api/v2/views/web_token.py
```diff
@@ -0,0 +1,88 @@
...
+        url = '%s?token=%s&password=display' % (
```
julianpistorius: Shouldn't need this dummy password bit I think. We should be able to add it in Nginx conf if necessary. Can't quite remember. I might just pull this branch & try it out.

lenards: We unfortunately do, I believe. This came up in the review of Guacamole Integration.

lenards: IIRC - we set a password for vncserver while configuring it via the instance deployment automation (aka `subspace` + `atmosphere-ansible`). I do not believe that the native VNC will let us skirt the password. But when I looked for the noVNC documentation on this, I came up empty. I think I can dig this _"detail"_ out of the ticket comments for this feature.

lenards: It looks like the `GET` parameter might be included to avoid having to input a password in the _hosting_ noVNC HTML:

https://github.com/novnc/noVNC/blob/master/vnc_auto.html#L225

julianpistorius: Correct. I remember trying to just make Nginx add this when proxying the connection. Should be possible. But I was at a conference, etc.

api/v2/views/web_token.py
steve-gregory: The reason for the FIXME hasn't quite been addressed (in review) so i'll ping @julianpistorius @lenards for comment:

- Should atmosphere be responsible for the URL generation _also_ or limited only to `token`?

julianpistorius: @steve-gregory Hmm.... From memory this comment of mine only referred to the fact that we shouldn't try to fingerprint the client's IP address.

lenards:
@steve-gregory - 

I think from looking at `urls.py`, I would expect that the endpoint `/v2/web_tokens/` is accepting an instance-alias (aka provider_alias of an instance) and not instance.id. Am I right?

lenards:
Nevermind, I answered my own question:

```python
     # .... 
     instance = self.get_queryset().filter(provider_alias=instance_id).first()
```

lenards:
Just to share a bit of how I imagine the pull request in Troposphere will look. 

It will use `requests` to make a "sub-request" out to `api/v2/web_token/:instance_alias` in a similar way to how _emulate_ in `troposphere.views.emulation` does:
```python
    r = requests.get(
        os.path.join(settings.API_SERVER, "api/v1/web_token/%s" % instance_alias),
        headers={'Authorization': 'Token %s' % auth_token, 
                 'Accept': 'application/json', 'Content-Type': 'application/json'})
```
source: https://github.com/cyverse/troposphere/blob/master/troposphere/views/emulation.py#L46-L54

lenards:
I think this is 👍 :shipit:

lenards:
I believe this can be merged first, without the pull request from troposphere. Or, I can build out a PR based on this.

lenards:
One point, with the variables shifting around - we will need to apply updates to the build-env (aka secrets) repositories too. 

- [ ] update build env repositories

steve-gregory:
@lenards , so far the _content_ of the variables has not changed, only the _placement_ of the variables with respect to atmosphere/troposphere, [see this clank PR for details.](https://github.com/cyverse/clank/pull/180)

PR #370, opened 2017-05-24
New users will be added to an allocation source when their account is…

amit4111989:
… created

## Description

When a new user is added to CyVerse, their account will be associated an Allocation Source with default values


steve-gregory:
@amit4111989 This will work, but we need to skip this scenario for Jetstream. 

For now we can just add an `if "jetstream" in settings.INSTALLED_APPS`

PR #373, opened 2017-05-30
Changing  `only_current` query for `start_date` to be less restrictive (for Instances)

steve-gregory:
Specifically, the only_current queries related to Instance.
In the future, we may move to less restrictive queries
_for all_ cloud resources, to avoid 'time sync' errors/issues


steve-gregory:
:shipit: 🚅

PR #378, opened 2017-06-01
Improve the scripts/add_new_provider experience

steve-gregory:
- By including a sample cloud_config that works for most new operators
- By including a sample DevStack config that will help onboarding
- Hide password input from console, make the password question similar to openrc file.
- Create a 'to_jsonfile' for existing providers
    (which can be used to create new providers)


steve-gregory:
Shameless self-:shipit:

PR #387, opened 2017-06-14
More Systemd changes

calvinmclean:
## Description

This pull request does a few things to improve the systemd files:
     - `flower` and `celeryd` are now independent. Previously, they were bound together so they would start and stop at the same time. 
     - Created a new service, `atmosphere-celery` that will control all Atmosphere processes
     - Now, `atmosphere` service manages only `nginx` and `uwsgi`
     - `redis-server` is now a soft dependency, or `Wants`, in Atmosphere services so it is not stopped and restarted by Atmosphere, but will be started by Atmosphere (and if `redis` fails to start, Atmosphere does not care)
     - Atmosphere services no longer list their parts (`nginx`, `uwsgi`, etc) as `Requires` or `After` but instead starts them in the `ExecStart` directive. This is more reliable and straight forward in my experience.
     - Stopping `uwsgi` is done better now, but requires changes in Clank. Previously, I couldn't figure out how `%(deb-confnamespace)` and other vars were processed in uwsgi default.ini. I discovered that initd uses environment variables to fill in these values so I added those to the systemd files. Changes made to previous Clank PR can be undone now.

## Related Pull Requests
https://github.com/cyverse/clank/pull/185

- [x] Documentation created/updated at `extras/systemd/README.md` to give context to the feature
- [ ] Clank PR coming soon


extras/systemd/atmosphere-full.service
```diff
@@ -0,0 +1,14 @@
+[Unit]
+Description=CyVerse Atmosphere Full: celery, celerybeat, uwsgi, nginx
```
steve-gregory: 👍

PR #389, opened 2017-06-15
Upgrade swift client version to support session

xuhang57:
## Description

This patch upgrades the swift client version to `3.3.0` so it supports session auth. 

`rtwo` related PR: https://github.com/cyverse/rtwo/pull/13


requirements.txt
```diff
@@ -100,7 +100,7 @@ python-logstash==0.4.5
 python-neutronclient==6.0.0  # via rtwo
 python-novaclient==6.0.0  # via python-openstackclient, rtwo
 python-openstackclient==3.3.0  # via rtwo
-python-swiftclient==3.1.0  # via rtwo
+python-swiftclient==3.3.0  # via rtwo
```
steve-gregory: A change will also be required to `rtwo==` in requirements after the `rtwo PR` is merged into the code base.

xuhang57: Thanks, done!

PR #398, opened 2017-06-22
Problem: Missing UserAllocationSources (and more!)

julianpistorius:
*Problem 1*

Users reported being unable to launch instances because they had no
 allocation sources. On investigation it turned that the
 `AllocationSources` with their names did exist, but not the
 `UserAllocationSources` which were used to link their `AtmosphereUser`
 accounts to the allocation sources.

We also found that the `user_allocation_source_created` events which
 should trigger the creation of the `UserAllocationSources` did exist.

The process for creating a `UserAllocationSource` relies on an
 `AtmosphereUser` already existing, and happens as part of creating an
 identity on a provider - in the
 `service.accounts.openstack_manager.AccountDriver#create_account`.
 Normally this happens when the user logs in and we know that the
 `AtmosphereUser` exists.

We found another another place where `create_account` is called.

The `add_new_accounts.py` script is called by a cron job, looks up
 new LDAP accounts with Atmosphere access, and creates identities on a
 provider using `create_account`. It doesn't create an `AtmosphereUser`
 before it calls `create_account`.

As a result `create_account` would fail when creating the
 `UserAllocationSource`, but not before it creates an `AllocationSource`
 matching the user name. Any calls to `create_account` also check if an
 `AllocationSource` exists with the same name as the username so that it
 does not create duplicate `AllocationSources`. It also does not try to
 create (the now missing) `UserAllocationSource`.

This is how we ended up with an `AllocationSource` for a username as
 well as an `user_allocation_source_created` event, but no
 `UserAllocationSource`.

Solution: In `add_new_accounts.py`, before calling `create_account`
 ensure that the AtmosphereUser exists by calling
 `core.models.Group.create_usergroup`

Also add a 'safety valve'/'eject button' in
 `service.accounts.openstack_manager.AccountDriver#create_account` to
 throw an exception if is called with a username that is not present
 in the system.

*Problem 2*

In `create_account` these allocation sources for usernames are also
 being created on Jetstream. This is not good. Fortunately they are
 being deleted as soon as they are created because of the TAS
 'mirroring' logic, but still...

Solution: Created an AllocationSourcePluginManager

There are two plugins to start with:

- `jetstream.plugins.allocation_source.JetstreamAllocationSource`
- `cyverse_allocation.plugins.allocation_source.FlexibleAllocationSource`

Currently the only required method for a plugin of this type is:
 `ensure_user_allocation_source` and it should return a boolean.

This will ensure that the user has a valid allocation source,
 potentially creating it in the case of the `FlexibleAllocationSource`.

I still have to implement the logic for `JetstreamAllocationSource`, and
 it just returns `True` for now.

- [ ] ~If necessary, include a snippet in CHANGELOG.md~

### DIFF
--- a/core/plugins.py
+++ b/core/plugins.py
@@ -64,6 +64,46 @@ class DefaultQuotaPluginManager(PluginManager):
         return _default_quota
 
 
+class AllocationSourcePluginManager(PluginManager):
+    """
+    Provide a plugin to create more complicated rules for default quotas
+    """
+    list_of_classes = getattr(settings, 'ALLOCATION_SOURCE_PLUGINS', [])
+    plugin_required = True  # For now...
+
+    @classmethod
+    def ensure_user_allocation_sources(cls, user, provider=None):
+        """Load each Allocation Source Plugin and call `plugin.ensure_user_allocation_source(user)`
+
+        Depending on the plugin this may create allocation sources if they don't already exist.
+        :param user: The user to check
+        :type user: core.models.AtmosphereUser
+        :param provider: The provider (optional, not used by all plugins)
+        :type provider: core.models.Provider
+        :return: Whether the user has valid allocation sources
+        :rtype: bool
+        """
+        _has_valid_allocation_sources = False
+        for AllocationSourcePlugin in cls.load_plugins(cls.list_of_classes):
+            plugin = AllocationSourcePlugin()
+            try:
+                inspect.getcallargs(
+                    getattr(plugin, 'ensure_user_allocation_source'),
+                    user=user, provider=provider)
+            except AttributeError:
+                logger.info(
+                    "Allocation Source plugin %s missing method 'ensure_user_allocation_source'",
+                    AllocationSourcePlugin)
+            except TypeError:
+                logger.info(
+                    "Allocation Source plugin %s does not accept kwargs `user` & `provider`",
+                    AllocationSourcePlugin)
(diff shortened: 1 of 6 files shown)


service/accounts/openstack_manager.py
```diff
@@ -231,24 +237,14 @@ def create_account(self, username, password=None, project_name=None,
...
-                logger.debug('No Allocation Source with name %s found. Creating a new Allocation Source...', allocation_source_name)
-                # Create Allocation Source for User
-                self._create_allocation_source(allocation_source_name)
-                # Add User to Allocation Source
-                self._assign_user_allocation_source(allocation_source_name,username)
-            else:
-                logger.debug('Allocation Source with name %s exists. Skipping creation step...' , allocation_source_name)
+            has_allocations = core.plugins.AllocationSourcePluginManager.ensure_user_allocation_sources(user)
+            if not has_allocations:
+                raise ValueError('User "{}" has no valid allocations'.format(user))
```
steve-gregory: Great work @julianpistorius 👍

core/plugins.py
```diff
@@ -64,6 +64,46 @@ def default_quota(cls, user, provider):
         return _default_quota
 
 
+class AllocationSourcePluginManager(PluginManager):
+    """
+    Provide a plugin to create more complicated rules for default quotas
+    """
```
steve-gregory: 🏆

PR #401, opened 2017-06-29
Problem: number of uWSGI processes not configurable

lenards:
## Description

This work makes the number of `processes` within the uWSGI ini configuration file configurable from `variables.ini`. The `ini` file is built during an execution of Clank.

This pull request supports work proposed as [cyverse/clank PR 189](https://github.com/cyverse/clank/pull/189).


lenards:
It looks like there is a module import failure in Travis:
- https://travis-ci.org/cyverse/atmosphere/builds/248214380#L1216

But the `./configure` call succeeded. I'm going to wait until I'm not rested to try to understand this.

steve-gregory:
See #408 for the rebased version of this PR.

PR #403, opened 2017-06-29
Bugfix: invalid HTTP Host header causes 500

steve-gregory:
## Solution:
 Include nginx snippet to block calls with an invalid host header.


steve-gregory:
Self-:shipit:

cdosborn:
Right on, was silently hoping this would be brought over.

PR #405, opened 2017-07-06
Problem: Launch instance script fails - no allocation source

julianpistorius:
The `scripts\launch_instances.py` file calls
`service.instance.launch_instance` - which now expects an allocation
source.

Solution: Look up the allocation source matching the username and pass
that along.

### DIFF
--- a/scripts/launch_instances.py
+++ b/scripts/launch_instances.py
@@ -12,7 +12,7 @@ import libcloud.security
 
 from django.db.models import Count, Q
 from core.models import AtmosphereUser as User
-from core.models import Provider, ProviderMachine, Size, InstanceSource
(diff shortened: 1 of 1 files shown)


scripts/launch_instances.py
```diff
@@ -184,7 +184,8 @@ def launch(user, name_prefix, provider, machines, size,
            host, skip_deploy, count):
     ident = user.identity_set.get(provider_id=provider.id)
     instances = []
-    kwargs = {}
+    allocation_source = AllocationSource.objects.get(name__exact=user.username)
```
steve-gregory: This fix won't work for _other sites_ but solves the problem for us, for now 👍

PR #408, opened 2017-07-06
Make number of uWSGI processes configurable

julianpistorius:
This work makes the number of `processes` within the uWSGI ini configuration file configurable from `variables.ini`. The `ini` file is built during an execution of Clank.

This pull request supports work proposed as [cyverse/clank PR 189](https://github.com/cyverse/clank/pull/189).


steve-gregory:
:shipit:

PR #410, opened 2017-07-10
Problem: Can't use Jetstream when TAS API is unreachable

julianpistorius:
Solution: Modified the `XsedeProjectRequired` validation plugin to have
a fallback strategy when the TAS API is unavailable.

The strategy is to check whether the user has existing allocation
sources in the database.

In order to achieve this I had to make a clear distinction between:

1. TAS API is unavailable or throwing bad errors
2. TAS API is available, and:
  a. There really is no TACC account for an XSEDE username
  b. We try to find allocations or projects for an invalid TACC account

We were conflating all of these before.

This means that existing Jetstream users can log in and use Atmosphere
as long as:

- They have logged in before
- They were valid the last time we were able to check with the TAS API

Note: New Jetstream users who have never logged into Atmosphere will be
disabled until we can check them against the TAS API at least once.

### DIFF
--- a/features/jetstream.features/login_tas_api_failover.feature
+++ b/features/jetstream.features/login_tas_api_failover.feature
@@ -1,0 +1,46 @@
+Feature: Should be able to use Jetstream, even when Atmosphere can't reach TAS API
+  As long as:
+  - They have logged in before
+  - They were valid the last time we were able to check with the TAS API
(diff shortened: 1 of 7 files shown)


julianpistorius:
Travis build is failing due to the Trusty box issue.

PR #411, opened 2017-07-10
Revert 'Not resetting allocation sources'

julianpistorius:
Reverting 4cb27801076b3fc8b04b6c492315e2cd8f25ec38 because it is not necessary.

```
Problem: Not resetting allocation sources

Solution: Re-use `cyverse_allocation.tasks.renew_allocation_sources`,
but simplified for a simple reset of the allocation source which does
not require the rules engine.

Next: Use the rules engine.
```

Turns out that `cyverse_allocation.tasks.update_snapshot_cyverse`
runs the rules engine which renews the allocation sources.


julianpistorius:
Tests are failing because of the manually edited `requirements.txt`

PR #414, opened 2017-07-10
Problem: Incorrect `pip-compile` command in instructions

julianpistorius:
Solution: Removed extra, incorrect file argument

- [x] Documentation created/updated at [Example link to documentation](https://example.test/doc#new_section) to give context to the feature

### DIFF
--- a/REQUIREMENTS.md
+++ b/REQUIREMENTS.md
@@ -32,7 +32,7 @@ pip-sync dev_requirements.txt
 pip-compile -o requirements.txt requirements.in
 
 # Generate new development requirements (depends on previous command!!)
-pip-compile -o dev_requirements.txt  dev_requirements.in requirements.txt
+pip-compile -o dev_requirements.txt dev_requirements.in requirements.txt
 ```
 
 **NOTE:** If you forget how a requirements file is generated, the instructions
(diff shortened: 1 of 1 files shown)


REQUIREMENTS.md
```diff
@@ -76,7 +76,7 @@ generated requirements.
 
 Your second command is actually more like:
 ```
-$ pip-compile -o requirements.txt requirements.txt requirements.in
+$ pip-compile -o requirements.txt requirements.in
```
cdosborn: The proposed change is incorrect. If this section is unclear, we should change it. In the context of lines 57- 82, I think line 79 makes sense.

julianpistorius: Oops. I'll have a closer look.

julianpistorius: If I understand this correctly this line is trying to say that:

`pip-compile -o requirements.txt requirements.in`

is equivalent to:

`pip-compile -o requirements.txt requirements.txt requirements.in`

And is _not_ saying that you should run it like this to update the requirements.txt.

cdosborn: Yeah

julianpistorius: Ah! Thanks. That makes sense. Sorry about the confusion.

julianpistorius:
Closing. #not-a-bug.

PR #418, opened 2017-07-12
application_sync_providers

c-mart:
## Description
This script performs a one-way synchronization of Applications (a.k.a. images)
from a master Provider to one or more replica Providers.

Also, `application_to_provider` is now modular (usable as a module as well as a script called interactively). Did quite a bit of refactoring of Glance API calls to be compatible with Glance v1 API (which is needed to create Glance images with specific UUID on old (Havana) Providers).

Other things I had to do in order for this to work in Atmosphere(0):
- Patch Glance v1 API server in old Havana clouds so that `irods` is a supported external source
  ```
    @staticmethod
    def _validate_source(source, req):
        if source:
            for scheme in ['s3', 'swift', 'http', 'rbd', 'sheepdog', 'cinder', 'irods']:
                if source.lower().startswith(scheme):
                    return source
            msg = _("External sourcing not supported for store %s") % source
            LOG.debug(msg)
            raise HTTPBadRequest(explanation=msg,
                                 request=req,
                                 content_type="text/plain")
  ```
- Grant my test iRODS account (`test-cmart`) ownership of `/iplant/home/atmo-images-workshop` (in addition to the other iRODS collections used by OpenStack clouds), so that it could copy data objects there

Here is an example script call that synchronizes applications from Marana Cloud to Tucson Cloud and Workshop Cloud (limited to Application ID 1420):
`# ./application_sync_providers.py --irods-conn irods://test-cmart:[redacted]@data.iplantcollaborative.org:1247/iplant  --limit-app-ids 1420 --irods-collections '{"7": "/iplant/home/openstack-images-marana-cloud", "4": "/iplant/home/atmo-images", "5": "/iplant/home/atmo-images-workshop"}' --glance-client-versions '{"4":1, "5":1}' 7 4 5`

### Testing
- [x] Test in local dev environment: find application that is on Marana Cloud but not others; sync image to other providers and launch it
- After team review:
  - [ ]  Test in production: request a new image on Marana Cloud, then sync image to other providers and launch it
  - [ ] Test using `application_to_provider` for its original use case of migrating images/apps from old clouds to Marana Cloud, ensure that still works

### Other
- [ ] ~New test(s) included to reproduce the bug/verify the feature~
- [x] Documentation created/updated [here](https://github.com/cyverse/atmosphere-guides/pull/25) to give context to the feature
- [ ] ~If necessary, include a snippet in CHANGELOG.md~ - This was already added for z-z?

- We want this script to be called via `cron` -- should Clank configure this for us on deployment, with cron jobs passed via `group_vars`? (There is also `cron_add_new_accounts.sh` which Clank could configure)


c-mart:
I was abusing git as a file synchronization tool between my local workstation and development environment so ah, there are a lot of commits -- I can squash these before merge if desired.

scripts/application_sync_providers.py
```diff
@@ -0,0 +1,190 @@
...
+
+
+def _parse_args():
+    parser = argparse.ArgumentParser(description=description, formatter_class=argparse.RawDescriptionHelpFormatter)
+    parser.add_argument("master_provider_id", type=int, help="Master provider ID")
+    parser.add_argument("replica_provider_id", type=int, nargs='+', help="Replica provider ID(s)")
+    # https://stackoverflow.com/questions/18608812/accepting-a-dictionary-as-an-argument-with-argparse-and-python
+    parser.add_argument("--glance-client-versions", type=json.loads,
+                        metavar='{\"1\": 2, \"2\": 1}',
+                        help="JSON specifying Glance client version for each provider ID")
```
steve-gregory: Can we derive this information based on something that we already have available? (For example: Legacy cloud providers use `ex_force_auth_version: 2.0_password` while newer cloud providers use `ex_force_auth_version: 3.x_password`

c-mart: Maybe. I found that that legacy (Havana) providers require Glance API V1 client in order to create an image with a specific UUID. Do we have that kind of information stored somewhere?

steve-gregory: Upon further review, this argument is fine as is 👍

steve-gregory:
@c-mart can you take care of the merge conflict and i'll :shipit:

c-mart:
@steve-gregory done.

steve-gregory:
:shipit:

c-mart:
Tested and works in production as expected, so far. Now to call it via cron. I'll have a separate PR to Clank so that we can define configuration of cron tasks in variables.yml.

PR #426, opened 2017-07-17
[ATMO-1916] Track redeploy and reboot in InstanceStatusHistory

steve-gregory:
## Problem
Instance does not keep track of user-initiated redeploy, reboot, and hard reboot via instance actions.

## Solution
Ensure that a new `Instance Status History` entry is created for each of the three previously mentioned states. 


steve-gregory:
Hold off on review until test cases can be fixed

steve-gregory:
Self-:shipit:

PR #427, opened 2017-07-17
Sanity test cases for the v2 API

steve-gregory:
## Description

There is a subset of tests that should be included in all /v2 API endpoint testing. One example is ensuring that `<key>` in  `/v2/<type_of_resources>/<key>` is invalid, null, or improper type that the application does not "crash" with a 500 error.

This PR can and should be improved prior to merge.

### DIFF
--- a/api/tests/v2/base.py
+++ b/api/tests/v2/base.py
@@ -1,0 +1,70 @@
+from django.core.urlresolvers import reverse
+
+from rest_framework.test import APIClient, APIRequestFactory, force_authenticate
+
+
+class APISanityTestCase(object):
+    url_route = None
+    user = None
+
+    def test_list_response_is_paginated(self):
+        """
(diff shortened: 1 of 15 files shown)


api/tests/v2/base.py
```diff
@@ -0,0 +1,48 @@
...
+        url = reverse(url_route, args=("1-2-3-4",))
+        response = client.get(url)
+        self.assertEquals(response.status_code, 404)
+
+        url = reverse(url_route, args=("deadbeef-dead-dead-dead-beefbeefbeef",))
+        response = client.get(url)
+        self.assertEquals(response.status_code, 404)
+
+    class Meta:
+        abstract = True
```
steve-gregory: These lines do nothing and will be removed before merge.

steve-gregory:
Self-:shipit:

PR #431, opened 2017-07-19
Trim old migrations, [0001...0084] -> 0001

cdosborn:
## Description

In order to build the project from a database dump, your dump can be no older
than zesty-zapdos.

### Why was this done?

This data migration had a bug:
core/migrations/0084_create_and_assign_allocationsource_to_users.py

The bug doesn't have a trivial fix, it inadvertently refers to a db model,
that is not django protected, every model must be referred to using
apps.get_model.

### How can I build the app from an older dump?

In order to build from an ealier db dump: load the old dump, checkout
the snapshot where 0084 was introduced then migrate. Your database is now
equivalent to 0001 in this snapshot. You should be able to run migrate on code
after this point.

Please describe your pull request. If it fixes a bug or resolves a feature request, be sure to link to that issue. When linking to an issue, please use `refs #...` in the description of the pull request.


cdosborn:
To test that this feature worked. I took an old db dump and migrated it to `0084` (using the commit where it was introduced) and created a new dump. This new dump represents the last version of a db dump that will work with the new code. Then i checked out this feature and ran ```./manage.py migrate``` with success.

cdosborn:
@steve-gregory any preference for the naming here of the migrations

steve-gregory:
@cdosborn nope, I'm for leaving the name as-is.

PR #441, opened 2017-07-25
--metadata-only option for application_to_provider

c-mart:
## Description

Migrates/updates image metadata only, not image data.

Useful for the case where a private application was previously migrated before all the image members on source provider were added as projects on the destination provider. Running this with --metadata-only afterward would add the remaining members to the image on the destination provider, allowing those members to launch instances from the application.


steve-gregory:
:shipit:

PR #445, opened 2017-07-26
Problem: Failing tests

julianpistorius:
`edit_instance.feature` was using a step to create an identity in a way
that would only work when using the correct setup for Jetstream & TAS

Solution:
1. Use the correct step,
2. Document it in the step,
3. Add an `assert` to catch it in the future, and
4. Skip all these scenarios if running in Jetstream distribution


steve-gregory:
Thank you @julianpistorius

PR #447, opened 2017-07-27
Optimize resource request serialization

cdosborn:
## Description
**Problem:**
API call to resource requests is too slow
**Solution:**
Prefetch data in the nested serializers, create less objects

See [documentation](https://github.com/cyverse/atmosphere/pull/447/files#diff-88333ba11114e014e87198d04680144d).

The API endpoint for 400 models went from making 3000 db queries to about 10 and from roughly 20s total request time to 1s. Many of our apis will experience similar speedups from these fixes. 

Even though the queries were improved by a factor of 300, the time was only improved by a factor of 20, so database optimization isn't everything—serialization is also costly. 

- [x] Documentation created/updated at [Example link to documentation](https://example.test/doc#new_section) to give context to the feature

### DIFF
--- a/api/v2/serializers/details/resource_request.py
+++ b/api/v2/serializers/details/resource_request.py
@@ -1,9 +1,9 @@
 from rest_framework import serializers
 from core.models import ResourceRequest
-from api.v2.serializers.summaries import UserSummarySerializer
-from api.v2.serializers.fields import (
-    IdentityRelatedField, StatusTypeRelatedField
+from api.v2.serializers.summaries import (
+    UserSummarySerializer, StatusTypeSummarySerializer
 )
+from api.v2.serializers.fields import IdentityRelatedField
 from api.v2.serializers.fields.base import UUIDHyperlinkedIdentityField
 
 
@@ -14,7 +14,7 @@ class ResourceRequestSerializer(serializers.HyperlinkedModelSerializer):
     )
     created_by = UserSummarySerializer(read_only=True)
     identity = IdentityRelatedField(source='membership.identity')
-    status = StatusTypeRelatedField(allow_null=True, required=False)
+    status = StatusTypeSummarySerializer(read_only=True)
 
     class Meta:
         model = ResourceRequest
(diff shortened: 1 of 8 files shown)


api/v2/serializers/fields/status_type.py
```diff
@@ -8,8 +8,7 @@ class StatusTypeRelatedField(serializers.RelatedField):
     def get_queryset(self):
         return StatusType.objects.all()
 
-    def to_representation(self, value):
-        status_type = StatusType.objects.get(pk=value.pk)
+    def to_representation(self, status_type):
         serializer = StatusTypeSummarySerializer(
```
cdosborn: The line on 13 is notorious. The resource request viewset doesnt use this field anymore. See the comment above.

api/v2/serializers/details/resource_request.py
```diff
@@ -14,7 +14,7 @@ class ResourceRequestSerializer(serializers.HyperlinkedModelSerializer):
     )
     created_by = UserSummarySerializer(read_only=True)
     identity = IdentityRelatedField(source='membership.identity')
-    status = StatusTypeRelatedField(allow_null=True, required=False)
```
cdosborn: Because `StatusTypeRelatedField` creates a serializer in it's custom `to_representation`, for 400 resources requests 400 serializers were constructed. This doubled the time contributed by this one field.

api/v2/serializers/fields/identity.py
```diff
@@ -7,9 +7,7 @@ class IdentityRelatedField(serializers.RelatedField):
     def get_queryset(self):
         return Identity.objects.all()
 
-    def to_representation(self, value):
-        queryset = self.get_queryset()
-        identity = queryset.get(pk=value.pk)
```
cdosborn: @steve-gregory, I'm guessing this change is going to break other code. Need to find out what is depending on this behavior.

cdosborn:
Lol i implemented a forever while loop.. one sec

steve-gregory:
:shipit: 🚅

PR #449, opened 2017-07-28
Allow 'SCRIPTS' to be passed to user_deploy

steve-gregory:
## Highlight Reel:
- Remove LibcloudDeployment
- Update atmosphere-ansible [(see companion PR)](https://github.com/cyverse/atmosphere-ansible/pull/96)

## Description

Allows users to deploy boot scripts and make use of `$ATMO_USER`
The previous method, LibcloudDeployment, had an error but was also being deprecated.
This solution does away with the LibcloudDeployment in favor of ansible.

- [ ] [Reviewed and approved companion PRs](https://github.com/cyverse/atmosphere-ansible/pull/96)


julianpistorius:
How would one go about testing this? I'm thinking Operation Sanity might be the place to do this.

PR #458, opened 2017-08-04
Introduce linting 💅

cdosborn:
## Description

Introducing linting into a project can be tricky. This script runs a strict variant on newly added files, and a lenient variant on modified files. This is a win for two reasons:
- New code will be  💅 
- Old code will gradually become 💅 

The linting tool is [prospector](https://prospector.landscape.io/en/master/). It has a lot of common sense defaults, and bundles pyflakes/pylint and others. It supports vulture as a plugin so we can nuke all our dead code.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,6 +46,7 @@
 
   script:
     - ./travis/check_properly_generated_requirements.sh
+    - ./travis/lint.sh
     - python manage.py test --keepdb
     - python manage.py behave --keepdb --tags ~@skip-if-${DISTRIBUTION}
     - python manage.py makemigrations --dry-run --check
--- a/dev_requirements.in
+++ b/dev_requirements.in
@@ -14,6 +14,7 @@ ipython
 mock
 newrelic
 pep8
+prospector
 wheel
 vcrpy
 json_delta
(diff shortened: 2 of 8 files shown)


cdosborn:
note to self: make this robust against non-existent sha's

xuhang57:
This is really neat 👍 

```
Check Information
=================
         Started: 2017-08-04 13:43:28.998086
        Finished: 2017-08-04 13:46:50.653449
      Time Taken: 201.66 seconds
       Formatter: grouped
        Profiles: default, no_doc_warnings, no_test_warnings, strictness_medium, strictness_high, strictness_veryhigh, no_member_warnings
      Strictness: None
  Libraries Used: flask, celery, django
       Tools Run: dodgy, mccabe, pep8, profile-validator, pyflakes, pylint
  Messages Found: 3317
```
So it runs 6 tools together with one command.

cdosborn:
I tacked on two commits, to make the generate requirements script more robust. 

@steve-gregory, I was bumping up against a lot of issues even with the least strict version of the linter. I think this will be a large improvement, but it comes with a caveat. Atmo developers need to be willing to clean up files they just touch or fix bugs in (or revise the rules that are applied via the linter). Next pr will get a more lenient set of rules, that we run over all files.

steve-gregory:
:shipit:

PR #461, opened 2017-08-08
Problem: Race conditions in django-cyverse-auth negatively affecting application

steve-gregory:
Solution: Update to django-cyverse-auth 1.1.0

Imports and methods have changed, so atmosphere needs to be updated accordingly.


steve-gregory:
Ignoring failed test case, which has been assigned to @julianpistorius to be fixed on `v27`

PR #470, opened 2017-08-14
Differentiate between inactive/end-dated

steve-gregory:
### End-Dated providers:
End-dated providers are not visible within the GUI/API (Archived)
To include end-dated providers in API responses use '?archived=True'
End-dated providers will throw ProviderNotActive exception.

### Inactive providers:
Inactive providers are visible with the GUI/API
Inactive providers _are not available for instance launch_ within the GUI/API
Inactive providers can still have drivers created.

### DIFF
--- a/core/models/provider.py
+++ b/core/models/provider.py
@@ -203,7 +203,23 @@ class Provider(models.Model):
     def get_type_name(self):
         return self.type.name
 
+    def is_current(self):
+        """
+        Is current when:
+        - No end date present
+        - End date is earlier than 'now'
+        """
+        if not self.end_date:
+            return True
+        now = timezone.now()
+        return self.end_date < now
+
     def is_active(self):
+        """
+        Is active when:
+        - No end date present OR end date is earlier than now
+        - active' == True
+        """
         if not self.active:
             return False
         if self.end_date:
--- a/core/models/provider.py
+++ b/core/models/provider.py
@@ -207,18 +207,20 @@ class Provider(models.Model):
         """
         Is current when:
         - No end date present
-        - End date is earlier than 'now'
+        - End date is later than 'now'
         """
-        if not self.end_date:
-            return True
         now = timezone.now()
-        return self.end_date < now
+        return (not self.end_date) or (self.end_date >= now)
 
     def is_active(self):
         """
         Is active when:
         - No end date present OR end date is earlier than now
         - active' == True
+
+        NOTE: Due to how the API handles end-dated providers
+        It is possible that `self.active` is True
+         and `self.is_active()` is False.
         """
         if not self.active:
             return False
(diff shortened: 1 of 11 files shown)


core/models/provider.py
```diff
@@ -203,7 +203,23 @@ def get_platform_name(self):
     def get_type_name(self):
         return self.type.name
 
+    def is_current(self):
+        """
+        Is current when:
+        - No end date present
+        - End date is earlier than 'now'
```
cdosborn: Do you mean end date is later than now?

core/models/provider.py
```diff
@@ -203,7 +203,23 @@ def get_platform_name(self):
...
+        """
+        if not self.end_date:
+            return True
+        now = timezone.now()
+        return self.end_date < now
```
cdosborn: This logic can be cleaned up, return a single boolean expression.

I think this should be:
```
return not self.end_date || self.end_date > now
```

core/models/provider.py
```diff
@@ -203,7 +203,23 @@ def get_platform_name(self):
...
+
     def is_active(self):
+        """
+        Is active when:
+        - No end date present OR end date is earlier than now
+        - active' == True
+        """
```
cdosborn: This needs to be fixed, its possible that `this.active == True` and `this.is_active() == False`.

core/query.py
```diff
@@ -187,7 +184,7 @@ def _in_range():
 
     if not now_time:
         now_time = timezone.now()
-    return _in_range() & _active_provider()
+    return _in_range() & current_provider()
```
cdosborn: Use `&&`, `&` does not short circuit
https://stackoverflow.com/a/3845032/1213041

cdosborn:
What is the use case for this pr?

core/query.py
cdosborn: We can ignore this since it is unrelated.

steve-gregory: I believe this is actually overuse of the `&` operator to create a `query & query` as the result returned.

core/models/provider.py
steve-gregory: an end-dated provider will always be inactive.. 😄

core/models/provider.py
steve-gregory: yes 👍

core/models/provider.py
steve-gregory: The documentation of the function has been updated accordingly 👍

core/query.py
cdosborn: oh hell i'm so wrong

julianpistorius:
Is this PR supposed to be go in to the release?

steve-gregory:
:shipit: because this needs to be part of the `v27` release to include our `deprecation strategy` for old cloud providers.

PR #475, opened 2017-08-17
Problem: Creating volume (for project) fails

julianpistorius:
Problem: Creating volume (for project) fails

URL: `/api/v2/project_volumes`

Error:

```
django.db.models.fields.related_descriptors:RelatedObjectDoesNotExist: Volume has no instance_source.
```

Background: The refactoring of the project-volumes relationship meant
that the many-to-many table doesn't exist anymore. The way the new code
was structured meant that a new volume was created, instead of setting
the `project` attribute on an existing volume.

We only caught it because the volume blew up due to a missing 
attribute - `instance_source`.

Solution: Added an explicit `create` method on `ProjectVolumeSerializer`
which sets the `project` attribute on the volume

This overrides the default creation behavior which only made sense in
the prior arrangement.

### DIFF
--- a/features/steps/api_with_persona_steps.py
+++ b/features/steps/api_with_persona_steps.py
@@ -1,9 +1,12 @@
+import collections
 import datetime
 import decimal
 import json
 import time
 import unittest
 import uuid
+from unittest import TestCase
+from unittest.util import safe_repr
 
 import dateutil.parser
 import django.test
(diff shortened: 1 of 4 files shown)


features/steps/api_with_persona_steps.py
```diff
@@ -336,6 +339,103 @@ def create_provider_machine(context):
...
+    response = client.post(url,
+                           {
+                               'name': project_name,
+                               'description': project_name,
+                               'owner': user.username
+                           })
+    context.persona['response'] = response
+
+
+@when('I create a volume with name "{volume_name}" and size {volume_size:d} using API')
```
lenards: Is there rationale for when `I` is used versus `we`?

julianpistorius: Yes. And I am not being consistent.

My idea:
- Use 'I' when an action is performed as a persona, with credentials, through the API
- Use 'we' when an action is performed by the wizard behind the curtain using magic & direct database manipulation

This was not a convention from the beginning, but emerged from the chaos.

julianpistorius: Potentially even better than 'we' is to phrase it in passive voice to make it even clearer that this not something done by an 'actor' in the system. And to state it as a fact, instead of an imperative. Something like this (fabricated) example:

```gherkin
When an image appears in the database
Then a corresponding machine will be created
When I refresh the page
Then I should see an image with this name in my list of images
```

julianpistorius: Thanks for pointing this out. This _We/I_ thing really bugs me and your input motivates me to go in and fix it once and for all.

julianpistorius: @lenards I made the wording more consistent while squashing and tidying things up. Ta!

lenards:
Thank you Julian!

lenards:
:shipit:

PR #483, opened 2017-08-22
Update requirements.txt

steve-gregory:
django-cyverse-auth==1.1.2 should be used for the `v27` release.


steve-gregory:
Verified travis.ci success prior to creating PR -- :shipit:

PR #498, opened 2017-08-29
Ensure visibility handled consistently for images

lenards:
## Description

This work ensures that we mark private images as "shared" (=> you + admin).

With changes made to handle group ownership, images that were "private" are considered "shared"; the explanation is that they're shared with a user (you) and the admin - always. So a private or shared with specific user image are both treated as "shared" within the glance catalog (I
think).

This is noted and captured here as a fix for issue related to imaging encountered during a verification of "private" images, and this updated semantic is required.

- [ ] ~New test(s) included to reproduce the bug/verify the feature~
- [ ] ~Documentation created/updated at [Example link to documentation](https://example.test/doc#new_section) to give context to the feature~
- [ ] ~If necessary, include a snippet in CHANGELOG.md~
- [ ] ~New variables supported in Clank~
- [ ] ~New variables committed to secrets repos~

### DIFF
--- a/service/machine.py
+++ b/service/machine.py
@@ -216,7 +216,7 @@ def upload_privacy_data(machine_request, new_machine):
 
     if is_public:
         print "Marking image %s private" % img.id
-        accounts.image_manager.update_image(img, visibility='private')
+        accounts.image_manager.update_image(img, visibility='shared')
 
     accounts.clear_cache()
     admin_driver = accounts.admin_driver  # cache has been cleared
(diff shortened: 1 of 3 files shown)


service/machine.py
```diff
@@ -216,7 +216,7 @@ def upload_privacy_data(machine_request, new_machine):
 
     if is_public:
         print "Marking image %s private" % img.id
```
julianpistorius: LOL

lenards: I think this is a matter of the _label_ `'private'` is no longer being used, anything `'private'` is not `'shared'`. The logic here seems confusing on face-value. 

But I think we mark "public" image private during the machine-request phase to avoid them showing up in glance-image catalog.

lenards:
ONWARD!

:shipit:

PR #500, opened 2017-09-01
Problem: New relic must be added after template configuration

steve-gregory:
Solution: Provide optional support for new relic via template

- [ ] [New variables supported in Clank](https://github.com/cyverse/clank/pull/208)

### DIFF
--- a/atmosphere/settings/local.py.j2
+++ b/atmosphere/settings/local.py.j2
@@ -397,6 +397,14 @@ if has_logstash and LOGSTASH_HOST:
                                      message_type='atmo-deploy', version=1)
     threepio.deploy_logger.addHandler(fh)
 
+
+# NEW RELIC
+try:
+    from atmosphere.settings import new_relic
+    NEW_RELIC_CONFIGURED = True
+except ImportError:
+    NEW_RELIC_CONFIGURED = False
+
 {% if TESTING %}
 {% include "atmosphere/settings/testing.j2" %}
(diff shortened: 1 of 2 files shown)


atmosphere/settings/local.py.j2
```diff
@@ -397,6 +397,14 @@ if has_logstash and LOGSTASH_HOST:
...
     threepio.deploy_logger.addHandler(fh)
 
+
+# NEW RELIC
+try:
+    from atmosphere.settings import new_relic
+    NEW_RELIC_CONFIGURED = True
+except ImportError:
+    NEW_RELIC_CONFIGURED = False
+
```
lenards: I think we'd need to bring the settings variables from `atmosphere.settings.new_relic` into the `atmosphere.settings` context here or in `atmosphere/settings/__init__.py`. 

Because for atmosphere, we use the `NEW_RELIC_ENVIRONMENT` string in the [`wsgi.py`](https://github.com/cyverse/atmosphere/blob/master/atmosphere/wsgi.py) start up code:
- https://github.com/cyverse/atmosphere/blob/master/atmosphere/wsgi.py#L38-L51

For Troposphere, the template to inject the browser-agent would also be in the `.new_relic`.

Maybe I'm misunderstanding the scoping for the Python namespace here.

steve-gregory:
:shipit: tested new more explicit call works and sets wsgi.py appropriately.

PR #503, opened 2017-09-05
Problem: OpenstackLoginBackend fails after new account creation logic

steve-gregory:
## Solution:
- Update the AccountCreationPlugin topology to better explain which plugin to use, and why.
- Continue with a simple approach that uses `/token_update` to do all of the heavy lifting
- Create a new 'AccountCreationPlugin' (DirectLogin) and include the ability to auto-create a new projects for the user  as part of the plugins settings.
- New settings available: `ALLOCATION_SOURCE_COMPUTE_ALLOWED`, `AUTO_CREATE_NEW_PROJECTS`

- [x] [New variables supported in Clank](https://github.com/cyverse/clank/pull/211)
- [x] New variables committed to secrets repos
- [ ] [Reviewed and approved the Troposphere Companion PR for Unlimited AU support.](https://github.com/cyverse/troposphere/pull/708)

### DIFF
--- a/api/tests/v2/test_allocation_sources.py
+++ b/api/tests/v2/test_allocation_sources.py
@@ -1,4 +1,6 @@
+import decimal
 import itertools
+import uuid
 from unittest import skip
 from django.core import urlresolvers
 from rest_framework.test import APIClient, APIRequestFactory
@@ -8,6 +10,7 @@ from api.tests.factories import (
     UserFactory, AnonymousUserFactory, IdentityFactory, ProviderFactory, AllocationSourceFactory,
     UserAllocationSourceFactory
 )
+from cyverse_allocation.tasks import update_snapshot_cyverse
 from .base import APISanityTestCase
 from api.v2.views import AllocationSourceViewSet as ViewSet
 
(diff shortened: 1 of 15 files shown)


steve-gregory:
## ToDo:
- [ ] GUI support for unlimited allocation when `compute_allowed == -1`

xuhang57:
so if I set `compute_allowed` == -1, users would have unlimited allocation for them to use right? And that actually depends on their OS quotas?

steve-gregory:
@xuhang57 this is unrelated to openstack quota, but will allow you to "disable/ignore" the time component in atmosphere (a.k.a. allocations).

If an instance didn't launch, the reason would be quota. We are working on a separate PR for Syncing OS quotas, both of which, combined, will get you the behavior you desire.

xuhang57:
Not sure wheater there is a point for backporting this in v27? Would v28 have lots of changes compared with v27? Thanks!

steve-gregory:
@xuhang57 this won't be backported to `v27` -- we're shooting for `v28` as the release for this feature. (Including unlimited allocation source support!)

xuhang57:
Make sense. Thanks!

steve-gregory:
:shipit:

PR #507, opened 2017-09-07
Support for django-prometheus

c-mart:
## Description

Adds support for [django-prometheus](https://github.com/korfuri/django-prometheus) which can be conditionally enabled by the deployer.

Exports Atmosphere metrics via the `/atmo-prometheus/metrics` URL.
Co-depends on [PR #214 in clank](https://github.com/cyverse/clank/pull/214).
Haven't yet added the [model monitoring mixins](https://github.com/korfuri/django-prometheus#monitoring-your-models) but we could.

  - [x] Tested in atmo-local to ensure Clank succesfully deploys and metrics are available at https://local.atmo.cloud/atmo-prometheus/metrics

### DIFF
--- a/atmosphere/settings/__init__.py.j2
+++ b/atmosphere/settings/__init__.py.j2
@@ -127,6 +127,9 @@ STATIC_URL = '/static/'
 ATMOSPHERE_NAMESPACE_UUID = UUID("40227dff-dedf-469c-a9f8-1953a7372ac1")
 
 MIDDLEWARE_CLASSES = (
+    {% if ENABLE_DJANGO_PROMETHEUS %}
+    'django_prometheus.middleware.PrometheusBeforeMiddleware',
+    {% endif %}
     'corsheaders.middleware.CorsMiddleware',
     # corsheaders.middleware.CorsMiddleware Must be ahead of
     # configuration CommonMiddleware for an edge case.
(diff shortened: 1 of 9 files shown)


julianpistorius:
Hmm... Would be kinda nice to specify the version of Clank required - so it doesn't break like this. Or some other way to manage the inter-project dependencies.

c-mart:
@julianpistorius if you don't use a version of Clank with [Clank PR #214](https://github.com/cyverse/clank/pull/214/files) then Atmosphere shouldn't break and things should work fine -- you just won't be able to deploy with django-prometheus. The default behavior here is not to set it up unless explicitly enabled in Clank.

atmosphere/settings/__init__.py.j2
```diff
@@ -139,6 +142,10 @@
     'django.contrib.messages.middleware.MessageMiddleware',
     'atmosphere.slash_middleware.RemoveSlashMiddleware',
     'atmosphere.slash_middleware.RemoveCSRFMiddleware',
+    {% if ENABLE_DJANGO_PROMETHEUS %}
+    'django_prometheus.middleware.PrometheusAfterMiddleware',
+    {% endif %}
+
```
steve-gregory: Lets not do this here -- you can inject django_prometheus to the beginning/end of `MIDDLEWARE_CLASSES` in `atmosphere/settings/local.py.j2` [here is an example of adding to the end](https://github.com/cyverse/atmosphere/blob/master/atmosphere/settings/local.py.j2#L440-L443)

PR #511, opened 2017-09-12
(WIP) Problem: Django tests fail when run consecutively

julianpistorius:
In particular failures happen in the VCR checks for whether the
cassettes have all been played. This is because the TAS API results are
cached on disk.

Solution: Disable caching when testing by using the `DummyCache` backend

### DIFF
--- a/atmosphere/settings/local.py.j2
+++ b/atmosphere/settings/local.py.j2
@@ -575,9 +575,17 @@ WEB_DESKTOP = {
     }
 }
 
+{% if TESTING %}
+CACHES = {
+    'default': {
+        'BACKEND': 'django.core.cache.backends.dummy.DummyCache',
+    }
+}
+{%- else %}
 CACHES = {
     'default': {
         'BACKEND': 'django.core.cache.backends.filebased.FileBasedCache',
         'LOCATION': '/var/tmp/django_cache',
     }
 }
+{%- endif %}


atmosphere/settings/local.py.j2
```diff
@@ -575,9 +575,17 @@ WEB_DESKTOP = {
     }
 }
 
+{% if TESTING %}
+CACHES = {
+    'default': {
+        'BACKEND': 'django.core.cache.backends.dummy.DummyCache',
+    }
+}
```
steve-gregory: This could lead to unintended consequences.. `TESTING` previously _enabled_ the ability to run `./manage.py test`. 

Rather than overloading `TESTING` perhaps we could go [this route](https://docs.djangoproject.com/en/1.11/topics/testing/tools/#overriding-settings) instead?

julianpistorius: Wouldn't be the first time I'm confused about the historical purpose for a variable/setting. The problem is we probably need caching disabled for _all_ tests - so it becomes very tedious very quickly. But maybe your idea of making the caching backend configurable is the answer.

julianpistorius:
Ok, I'll rework this PR to make the backend configurable instead.

cdosborn:
Closed in favor of #626

PR #515, opened 2017-09-13
V27 fix broken jetstream admin quota update

cdosborn:
## Description
This is a hack.

Problem:
In j7m, users don't make resource requests against specific
providers, yet each request is tied to a provider (identity). So admins in j7m
are approving quota essentially for one (random) identity of the user's identities.

Solution:
In the action for changing quota (specific to resource requests)
update the quota for all the user's identities.

### DIFF
--- a/api/v2/serializers/details/resource_request_actions.py
+++ b/api/v2/serializers/details/resource_request_actions.py
@@ -39,17 +39,20 @@ class ResourceRequest_UpdateQuotaSerializer(serializers.Serializer):
         user = validated_data.get('approved_by')
         resource_request = validated_data.get('resource_request')
         identity = validated_data.get('identity')
-        data = {
-            'identity': identity.id,
-            'resource_request': resource_request.id,
-            'approved_by': user.username,
-            'quota': quota.id
-        }
-        event_serializer = QuotaAssignedByResourceRequestSerializer(data=data)
-        event_serializer.is_valid(raise_exception=True)
 
-        # Synchronous call to EventTable -> Set Quota for Identity's CloudProvider -> Save the Quota to Identity
-        event_serializer.save()
+        # This api is designed to update a single identity's quota, instead we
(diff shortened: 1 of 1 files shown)


api/v2/serializers/details/resource_request_actions.py
```diff
@@ -39,17 +39,20 @@ def create(self, validated_data):
...
-        data = {
-            'identity': identity.id,
-            'resource_request': resource_request.id,
-            'approved_by': user.username,
-            'quota': quota.id
-        }
-        event_serializer = QuotaAssignedByResourceRequestSerializer(data=data)
-        event_serializer.is_valid(raise_exception=True)
 
-        # Synchronous call to EventTable -> Set Quota for Identity's CloudProvider -> Save the Quota to Identity
```
steve-gregory: Might want to move this comment with the block.

api/v2/serializers/details/resource_request_actions.py
```diff
@@ -39,17 +39,20 @@ def create(self, validated_data):
...
-        event_serializer.save()
+        # This api is designed to update a single identity's quota, instead we
+        # update all identities for the user. This syncs the quota across all of
+        # the user's active identities.
+        for ident in [i for i in user.identity_set.all() if i.is_active()]:
```
steve-gregory: Depending on the length of `user.identity_set.all()` this could be a long/expensive call. 

Moving this entire block into an AsyncTask wouldn't be unreasonable.

api/v2/serializers/details/resource_request_actions.py
cdosborn: I nuked it, because its just saying what the code does.

api/v2/serializers/details/resource_request_actions.py
cdosborn: Yeah we will need to do that.

PR #518, opened 2017-09-19
Quickfix for setting reclaim_ips

steve-gregory:
This is a HACK!
- Set to False if the provider is 'iPlant Cloud - Tucson'
- all other providers set to True (The default operation for all clouds)

### DIFF
--- a/service/instance.py
+++ b/service/instance.py
@@ -2009,7 +2009,11 @@ def run_instance_action(user, identity, instance_id, action_type, action_params)
     # Gather instance related parameters
     provider_uuid = identity.provider.uuid
     identity_uuid = identity.uuid
+
+    # NOTE: This if statement is a HACK! It will be removed when IP management is enabled in an upcoming version. -SG
     reclaim_ip = True if identity.provider.location != 'iPlant Cloud - Tucson' else False
+    # ENDNOTE
+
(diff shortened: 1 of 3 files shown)


service/instance.py
```diff
@@ -2032,25 +2037,25 @@ def run_instance_action(user, identity, instance_id, action_type, action_params)
...
                                      provider_uuid, identity_uuid,
-                                     user)
+                                     user, reclaim_ip)
     elif 'unshelve' == action_type:
         result_obj = unshelve_instance(esh_driver, esh_instance,
                                        provider_uuid, identity_uuid,
                                        user)
     elif 'shelve_offload' == action_type:
-        result_obj = offload_instance(esh_driver, esh_instance)
+        result_obj = offload_instance(esh_driver, esh_instance, reclaim_ip)
```
julianpistorius: Problem.

julianpistorius: See how it's called here:

https://github.com/cyverse/atmosphere/blob/7778e1e7ae8e781d4689af14110b4285247a4976/service/monitoring.py#L178

Also should we add `reclaim_ip` to this occurrence?

steve-gregory: @julianpistorius Yes, ill address those shortly thanks for bringing it to my attn

steve-gregory:
:shipit: after comments made during review were addressed and travis.ci was ✔️ . Thanks @julianpistorius

PR #519, opened 2017-09-19
Problem: Race condition can create two fixed IPs, causing networking errors

steve-gregory:
## Solution
- Given multiple ports, select a port which is ACTIVE and attached
  - Fix provided in the 'ready_to_deploy' task, so that it can be caught
    and fixed with a simple redeploy.

## Additional/semi-related changes
- Rename task: 'add_floating_ip'->'initialize_instance_networking'
  Rationale:
  - This task is doing a _whole lot_ more than just adding a floating IP
  - Goal of the task: instance is ready to network by the end of the task.

- Set metadata to 'initializing' for a redeploy/reboot
  Rationale:
  -  avoid 'bouncing back' to error states

- [x] Tested and verified on "broken" instances and new instances

### DIFF
--- a/service/instance.py
+++ b/service/instance.py
@@ -437,11 +437,25 @@ def redeploy_instance(
 
     if type(esh_driver) == AtmosphereMockDriver:
         return
+    # HACK: Forces a metadata update to avoid "Instance activity workflow" errors in the GUI
+    # by setting 'tmp_status' to 'initializing' users will not be stuck in a "final state"
+    # if API polling starts _prior_ to instance action being successful.
+    # When 'Instance activity workflow' has been addressed, remove these lines.
+    metadata = {'tmp_status': 'initializing'}
+    _update_instance_metadata(
+        esh_driver,
+        esh_instance,
+        metadata
+    )
+    esh_instance = esh_driver.get_instance(esh_instance.id)
+    # END-HACK
+
     # Start deployment chain based on the instance above
     deploy_chain = get_idempotent_deploy_chain(
         esh_driver.__class__, esh_driver.provider, esh_driver.identity,
         esh_instance, core_identity, core_identity.created_by.username)
     deploy_chain.apply_async()
+    return
 
 
 def restore_ip_chain(esh_driver, esh_instance, redeploy=False,
@@ -1765,7 +1779,7 @@ def update_instance_metadata(core_instance, data={}, replace=False):
     esh_instance = esh_driver.get_instance(instance_id)
     return _update_instance_metadata(esh_driver, esh_instance, data, replace)
 
-def _update_instance_metadata(esh_driver, esh_instance, data={}, replace=True):
+def _update_instance_metadata(esh_driver, esh_instance, data={}, replace=False):
     """
     NOTE: This will NOT WORK for TAGS until openstack
     allows JSONArrays as values for metadata!
--- a/service/tasks/driver.py
+++ b/service/tasks/driver.py
@@ -77,7 +77,7 @@ def complete_resize(driverCls, provider, identity, instance_alias,
         driver = get_driver(driverCls, provider, identity)
         instance = driver.get_instance(instance_alias)
         if not instance:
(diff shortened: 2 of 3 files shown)


service/instance.py
```diff
@@ -442,11 +442,21 @@ def redeploy_instance(
 
     if type(esh_driver) == AtmosphereMockDriver:
         return
+    # Force a metadata update to avoid reverting to previous state.
```
cdosborn: On first blush this doesn't seem right. get_idempotent_deploy_chain works by looking at the current state, and acting. So no matter what state the instance is in (before your call here to set "initializing"), get_idempotent_deploy should be able to act accordingly right? 
If you look at this code, you'd ask yourself, "why would we force to avoid reverting?", because its not obvious within the context of *this* function. This change only makes sense when you know how another chunk of the code works in `get_deployment_chain`, which is a hint that code is being changed in the wrong place.

service/tasks/driver.py
```diff
@@ -881,6 +881,29 @@ def deploy_ready_test(driverCls, provider, identity, instance_id,
...
+            instance_floating_ips = [fip for fip in floating_ips if fip.get("instance_id",'') == instance_id]
+            selected_floating_ip = instance_floating_ips[0] if instance_floating_ips else None
+            if selected_floating_ip and port_id:
+                floating_ip_addr = selected_floating_ip["floating_ip_address"]
+                previous_port_id = selected_floating_ip["port_id"]
+                if previous_port_id != port_id:
+                    celery_logger.info(
+                            "Re-setting existing floating_ip_address %s port: %s -> %s",
+                        floating_ip_addr, previous_port_id, port_id)
+                    network_driver.neutron.update_floatingip(selected_floating_ip['id'], {'floatingip': {'port_id': port_id}})
```
cdosborn: Should deploy ready test be thought of as a question:"are we ready", or should it be: "ensure that we're ready"

service/tasks/driver.py
```diff
@@ -847,11 +847,10 @@ def _deploy_ready_failed_email_test(
 
 @task(name="deploy_ready_test",
       default_retry_delay=64,
-      # 16 second hard-set time limit. (NOTE:TOO LONG? -SG)
       soft_time_limit=120,
-      max_retries=300  # Attempt up to two hours
+      max_retries=115  # Attempt up to two hours
```
cdosborn: Could you comment on these time values? Confused about the range of possibilities this enables, like whats the slowest each could take * 115

cdosborn: Is the comment wrong?

service/instance.py
steve-gregory: You are probably right, that this change is happening in the wrong place.

The "Problem scenario" is:
- GUI (via user click) requests `reboot/hard reboot/redeploy`
- API forwards request to cloud, which accepts
- GUI begins Instance Polling
- API returns the last known instance state (In some cases, prior to the change being executed)
- GUI 'reverts' state back to a "known final state" (active, deploy_error, etc.)
- GUI never polls again, so user thinks this is the "end of the line"

- On page refresh//next poll, the action will have happened, and the appropriate status will be updated.

This is an *improvement* on the current behavior, in that it will avoid the 'final state, stop polling' step.. But it is not a *fix* for the larger bug described.. (I would be fine with writing up a JIRA to be addressed later, though..)

service/tasks/driver.py
steve-gregory: `deploy_ready_test` was traditionally designed to be an `are we ready` test.

The *bugfix* in question was applied here because it gives users the fastest path back to success (Redeploy, reboot, and your instance will be "fixed")

The code was originally placed in `add_floating_ip/initialize_instance_networking`, but when the instance already has a floating IP (albeit incorrectly attached) that task is not executed.

I'm all 👂 if theres a better way, though..

service/tasks/driver.py
julianpistorius: Looks like the comment used to be wrong when `max_retries` was 300. It's now correct. :)

PR #523, opened 2017-09-26
WIP - Simplify atmosphere-ansible variables

cdosborn:
## Description

Problem: Same variable is defined in two places, easy to misconfigure
Solution: Merge the two, by atmosphere providing its variable

atmosphere-ansible is now less dependent on a correctly configured secrets
repo. Atmosphere is calling subspace with variables, which removes the need
for SSHKEYS to exist in atmosphere-ansible. It also means that clank no longer
has to template out an ssh config file.

See accompanying pr https://github.com/cyverse/atmosphere-ansible/pull/116

### DIFF
--- a/service/deploy.py
+++ b/service/deploy.py
@@ -68,9 +68,9 @@ def ansible_deployment(
         shared_users.append(username)
     extra_vars.update({
         "SHARED_USERS": shared_users,
-    })
-    extra_vars.update({
         "ATMOUSERNAME": username,
+        "ATMOSPHERE_KEYPAIR_FILE": settings.ATMOSPHERE_KEYPAIR_FILE,
+        "ATMOSPHERE_PRIVATE_KEYFILE": settings.ATMOSPHERE_PRIVATE_KEYFILE
     })
     pbs = execute_playbooks(
         playbooks_dir, host_file, extra_vars, limit_hosts,


service/deploy.py
```diff
@@ -68,10 +68,12 @@ def ansible_deployment(
...
     extra_vars.update({
         "SHARED_USERS": shared_users,
-    })
-    extra_vars.update({
         "ATMOUSERNAME": username,
     })
+    with open(settings.ATMOSPHERE_KEYPAIR_FILE, 'r') as f:
+        extra_vars.update({
+            "SSHKEYS": [ f.read() ],
+        })
```
cdosborn: Not sure we want to make this change. But if we did, we could remove SSHKEYS from group_vars/all in our secrets repos.

steve-gregory: I like this idea! In most situations, these values will be ==, but when they aren't it can result in locking yourself out of a box that you once had access to.

 What do you think, @c-mart?

c-mart: If atmosphere can SSH into the instance to use Ansible and deploy SSH keys to `authorized_keys`, then by definition, won't this SSH key already be present in `authorized_keys` (injected by nova/cloud-init)? What will this actually accomplish?

cdosborn: Right so atmosphere-ansible initially has ssh access via several methods (centos/ubuntu/root). The code above kick starts atmo-ansible to provide standardized access to atmosphere as the root user. To do that requires templating the authorized keys with a public key. This change makes it so that the key comes from atmosphere, not from group_vars in the secrets repo (should be the same key) just requires a vars creator to make sure they match. Another issue is that atmo-ansible creates a key if you don't provide a key, this created key will not exist in atmosphere though.. So i'd like the key to come from one place for both projects.

c-mart: Yep.. came to me when I woke up this morning. [atmo-ssh-setup](https://github.com/cyverse/atmosphere-ansible/tree/master/ansible/roles/atmo-ssh-setup) role that Calvin built a few months ago. Two things: 

1. If `SSHKEYS` is removed from `atmosphere-ansible/group_vars/all`, then when running [atmosphere-ansible manually with `ansible-playbook` for troubleshooting purposes](https://wiki.cyverse.org/wiki/display/csmgmt/Ansible+Deployment#AnsibleDeployment-RunningAnsiblemanuallyonAtmoorAtmobeta) (warning, that page is ancient), we would now also need to specify something like `-e SSHKEYS=['ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQCoKZzT+ztjWRiGf6d+MrRaY9+I8hOU9+GJS9ujrbwTY/OGQsMgH1EHISWZX5La6mAJryXkBO9qpfkS4nXAiDpWSPSCeGfF283AgnEXZw1xyR+j0XxvO4ysZlmF6K6dnlNzXLNWQF7I41Qd1pf6VHvNBVBKSbX6cXpAplLMMwvL6trALQZciS0XHBvA4LUKDvEuha79hoOPodlZlX3c8NRdlTczQrKH+MmWLZn+ETuSxoWygRphX3nz0HFThN0YdGIGmE0X2ZcIdStQ3T1HL/eRhUlTxCBXzq74uvNfwhN106DMadUejymb1MkjdQ7f7h2OjTQ/Cxny32ly5/DHnT6D root@my-atmosphere']` on the command line. So, it might be good to continue specifying this key in group_vars, but then we have more parallel data structures. (Perhaps we need an easy way to ask Atmosphere "which extra vars would you pass when deploying this particular instance?", which can then be pasted into the `ansible-playbook` command.)

2. If you look in our environment-specific repos, we're not only using the `SSHKEYS` variable to set up root access for the atmosphere server, but also for all of us on the team individually (our public SSH keys are all in that list). Specifying only one key here would mean that all of our individual keys don't get deployed to the root user's `authorized_keys` on the instance. I for one am in favor of that, since we all have the Atmosphere server's SSH keypair and can just use that when accessing instances for troubleshooting purposes. We should decide what the intention is and possibly remove all of those keys from our environment-specific repos as companion PRs to this one.

cdosborn: > 1.
I agree. I don't want to make calling atmopshere-ansible more painful, and this pr does increase the friction. A way to address that issue in addition to the issue raised by this pr (that there is redundancy), would be for clank to template out group_vars/all. Ansible allows you to make group_vars/all/ a directory and drop multiple yml files there. So we could have clank template out this, and copy over everything else in the secrets folder

> 2.
I was not aware of that. Intention behind this pr isn't to remove functionality. If no one has any strong opinions ill restore the existing behavior.

c-mart: @cdosborn 
1. That would be really sweet!
2. After thinking about it, I'm in favor of removing our personal SSH keypairs from Atmosphere instances, and I'm happy for this PR to do it. If I were an Atmosphere user and I saw a long list of keys that look like `ssh-rsa [blablabla] cdosborn@maggot` in root's `authorized_keys` file on my instance, I would be really suspicious. "Who are all these people and why do they have access to all of my stuff?" Just one key labeled "atmo" would not raise such suspicion.. "Oh, that must be how they provision my instance".

cdosborn: cdosborn@i-see-all-your-data-when-want-to

cdosborn:
I'm going to close this and implement the [suggested changes](https://github.com/cyverse/atmosphere/pull/523#discussion_r141158217) in a clank pr.

PR #524, opened 2017-09-29
Avoid global subspace import

cdosborn:
Problem: Top level import of subspace.runner creates race condition
Solution: Move calls into tasks that need subspace 

service/deploy.py gets evaluated by celery workers running as root or www-data. The import of subspace results in some side effects, ansible goes and creates directories (with the perms of the current user).

So if celery's imaging worker which runs as root imports ansible, then those
directories will have root permissions, breaking ansible for www-data (i.e.
the other celery workers).

### DIFF
--- a/service/deploy.py
+++ b/service/deploy.py
@@ -422,15 +422,15 @@ def configure_ansible(debug=False):
     """
     Configure ansible to work with service.ansible and subspace.
     """
-    subspace.set_constants("HOST_KEY_CHECKING", False)
-    subspace.set_constants(
-        "DEFAULT_ROLES_PATH", settings.ANSIBLE_ROLES_PATH)
+
     os.environ["ANSIBLE_DEBUG"] = "true" if debug else "false"
     if settings.ANSIBLE_CONFIG_FILE:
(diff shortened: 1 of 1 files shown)


service/deploy.py
```diff
@@ -421,15 +422,15 @@ def configure_ansible(debug=False):
...
     os.environ["ANSIBLE_DEBUG"] = "true" if debug else "false"
     if settings.ANSIBLE_CONFIG_FILE:
         os.environ["ANSIBLE_CONFIG"] = settings.ANSIBLE_CONFIG_FILE
         os.environ["PYTHONOPTIMIZE"] = "1" #NOTE: Required to run ansible2 + celery + prefork concurrency
-        # Alternatively set this in ansible.cfg: debug = true
-        subspace.constants.reload_config()
+    subspace.configure({
+        "HOST_KEY_CHECKING": False,
+        "DEFAULT_ROLES_PATH": [ settings.ANSIBLE_ROLES_PATH ]
+    })
```
steve-gregory: 👍 Thanks @cdosborn !

steve-gregory:
ToDo: 

- [ ] Update the version of subspace added to `master` in `requirements.txt`

PR #525, opened 2017-10-03
Optimize SQL queries to avoid slow /v2/images calls.

steve-gregory:
- when using 'id in' queries, use a list() over a QuerySet.
- *Most important* Rather than join membership tables, do a lookup on individual tables, and look for results in query.
- Use prefetch_related/select_related on API get_queryset call


steve-gregory:
Self-:shipit: for ⏲

PR #527, opened 2017-10-05
Fix trivial key mismatch

cdosborn:
## Description
Before atmosphere runs subspace, it checks that ssh deploy key is the key that
was injected into the instance. The check was failing because of a trailing
new line. Same key is used to inject as is to perform the check, just reading
from python didn't strip the newline.## Checklist before merging Pull Requests


steve-gregory:
Thanks @cdosborn

PR #531, opened 2017-10-12
Problem: machine monitoring creates errors in image catalog

steve-gregory:
Problem(s):
- (1) Application with empty access-list returned _all users_ which made private images "Shared with >128 Users" error show up
- (2) New applications are being created with 'v. VersionName' in the
- (3) New applications are being created for each version update.

Solution(s):
- (1) return AtmosphereUser.objects.none() when access_list is empty.
- (2) Imaging will include version name, and monitoring should strip
      version name before processing.
      (This was noted in comment but never implemented)
- (3) Rely on application_uuid metadata and "fall back" on the
      glance image name to be the secondary 'key' of truth.
      If API users are creating images with the _same name_ and _same
      application UUID_ then we will need to find a new way
      to address this problem.

Cleanup related to the solution:
- Do away with confusing logic about REPLICATION_DRIVER.
  the code replaces the acting account driver with
  _an account driver pointed at a different provider_
  which makes it hard to use in practice.
- Remove (most) HACK statements
- Re-factor logic for retrieving the 'entire access list'
- small changes to speed up monitoring
  by making less calls to '.get(image_id)'
  and catch spurious error when image is not found.
- Update documentation and filter statements name.

- [ ] Check with support team(s) to determine if this should be hot-fixed or wait until v29.

### DIFF
--- a/core/models/application.py
+++ b/core/models/application.py
@@ -51,12 +51,12 @@ class Application(models.Model):
         """
         from core.models import AtmosphereUser
 
-        if not self.access_list.count():
-            return AtmosphereUser.objects.none()
-
-        all_users = AtmosphereUser.objects.all()
+        all_users = AtmosphereUser.objects.none()
         for pattern_match in self.access_list.all():
-            all_users &= pattern_match.validate_users()
+            if pattern_match.allow_access:
+                all_users |= pattern_match.validate_users()
+            else:
+                all_users &= pattern_match.validate_users()
         return all_users
 
     @property
--- a/core/models/application.py
+++ b/core/models/application.py
@@ -50,6 +50,10 @@ class Application(models.Model):
         - Returns a list of Users who passed the test
         """
         from core.models import AtmosphereUser
+
+        if not self.access_list.count():
+            return AtmosphereUser.objects.none()
+
         all_users = AtmosphereUser.objects.all()
         for pattern_match in self.access_list.all():
             all_users &= pattern_match.validate_users()
@@ -473,14 +477,18 @@ def create_application(
     Build information (Based on MachineRequest or API inputs..)
     and RETURN Application!!
     """
-    new_app = None
+    application = None
 
     if not uuid:
         uuid = _generate_app_uuid(identifier)
 
-    existing_app = Application.objects.filter(uuid=uuid)
-    if existing_app.count():
-        new_app = existing_app[0]
+    uuid_match = Application.objects.filter(uuid=uuid).first()
+    if name and uuid_match and name == uuid_match.name:
+        # This is our "Best effort" at keeping 'copycat' images
+        # inside the image catalog, made by non-atmosphere (API) users
+        # due to the metadata/property transfer that occurs.
+        # _Hopefully_ copycat images will have a different name.
+        application = uuid_match
 
     if not name:
         name = "Imported App: %s" % identifier
@@ -504,15 +512,15 @@ def create_application(
             tags = [t.strip() for t in tags.split(',')]
         else:
             tags = [tags]
-    if new_app:
-        new_app.name = name
-        new_app.description = description
-        new_app.created_by = created_by_identity.created_by
-        new_app.created_by_identity = created_by_identity
-        new_app.private = private
-        new_app.save()
+    if application:
+        application.name = name
+        application.description = description
+        application.created_by = created_by_identity.created_by
+        application.created_by_identity = created_by_identity
+        application.private = private
+        application.save()
     else:
-        new_app = Application.objects.create(
+        application = Application.objects.create(
             name=name,
             description=description,
             created_by=created_by_identity.created_by,
@@ -520,11 +528,11 @@ def create_application(
             private=private,
             uuid=uuid)
     if tags:
-        updateTags(new_app, tags, created_by_identity.created_by)
+        updateTags(application, tags, created_by_identity.created_by)
     if access_list:
         for pattern_match in access_list:
-            new_app.access_list.add(pattern_match)
-    return new_app
+            application.access_list.add(pattern_match)
+    return application
 
 #FIXME: This class marked for removal
 class ApplicationScore(models.Model):
--- a/core/models/application.py
+++ b/core/models/application.py
@@ -358,17 +358,17 @@ class ApplicationMembership(models.Model):
         unique_together = ('application', 'group')
 
 
-def _get_owner_identity(provider, created_by, uuid_match):
-    if uuid_match:
-        #Use existing authorship on UUID matches.
-        # This will avoid bounceback on replica-machines
-        return uuid_match.created_by_identity
-    elif created_by:
+def _get_owner_identity(provider, created_by=None):
+    """
+    This private method is intenteded to ensure that a valid identity exists
+    for the 'owner' of an Application.
+    If the identity cannot be found, the AccountProvider will be the author.
+    """
+    if created_by:
         return _user_identity_lookup(
             str(provider.uuid),
             created_by.username)
     else:
-        logger.error("ERROR: No identity was found for image <%s> -- the 'author' will be admin" % identifier)
         return provider.get_admin_identity()
 
 
@@ -505,8 +505,12 @@ def create_application(
         name = "Imported App: %s" % identifier
     if not description:
         description = "Imported Application - %s" % name
-    if not created_by_identity:
-        created_by_identity = _get_owner_identity(provider, created_by, uuid_match)
+    if not created_by_identity and uuid_match:
+        #Use existing authorship on UUID matches.
+        # This will avoid bounceback on replica-machines
+        return uuid_match.created_by_identity
+    elif not created_by_identity:
+        created_by_identity = _get_owner_identity(provider, created_by)
 
     if not tags:
         tags = []
--- a/core/models/application.py
+++ b/core/models/application.py
@@ -360,7 +360,7 @@ class ApplicationMembership(models.Model):
 
 def _get_owner_identity(provider, created_by=None):
     """
-    This private method is intenteded to ensure that a valid identity exists
+    This private method is intended to ensure that a valid identity exists
     for the 'owner' of an Application.
     If the identity cannot be found, the AccountProvider will be the author.
     """
--- a/core/models/application.py
+++ b/core/models/application.py
@@ -358,6 +358,20 @@ class ApplicationMembership(models.Model):
         unique_together = ('application', 'group')
 
 
+def _get_owner_identity(provider, created_by, uuid_match):
+    if uuid_match:
+        #Use existing authorship on UUID matches.
+        # This will avoid bounceback on replica-machines
+        return uuid_match.created_by_identity
+    elif created_by:
+        return _user_identity_lookup(
+            str(provider.uuid),
+            created_by.username)
+    else:
+        logger.error("ERROR: No identity was found for image <%s> -- the 'author' will be admin" % identifier)
+        return provider.get_admin_identity()
+
+
 def _get_app_by_name(provider_uuid, name):
     """
     Retrieve app by name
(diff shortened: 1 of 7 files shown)


core/models/machine.py
```diff
@@ -62,20 +62,21 @@ def test_existence(cls, provider, identifier):
...
+        elif len(split_list) > 2:
             logger.warn(
                 "Version separator(%s) is ambiguous: %s"
                 % (version_sep, machine_name))
-            version_parts = machine_name.rpartition(version_sep)
-            split_list = [version_parts[0].trim(), version_parts[2].trim()]
+            version_parts = [
+                s.strip() for s in
+                machine_name.rpartition(version_sep)]
+            split_list = [version_parts[0], version_parts[2]]
```
cdosborn: These 10 or so lines of string splitting can be avoided by a dedicated version field in the metadata. Code is hard to verify and a name should not be parsed for semantic meaning (because a names purpose is to label not to define its parts (thats what data is for), all names should be valid, and a user expects this, changing the name should not break atmo, changing the metadata could)

core/models/machine.py
```diff
@@ -283,7 +284,15 @@ def convert_glance_image(glance_image, provider_uuid, owner=None):
     from core.models import AtmosphereUser
     image_id = glance_image.id
     machine_name = glance_image.name
-    application_name = machine_name  # Future: application_name will partition at the 'Application Version separator'.. and pass the version_name to create_version
+    default_version_name = '1.0'  # could be configurable if controversial
+    application_name, version_name = ProviderMachine._split_cloud_name(machine_name)
```
cdosborn: See comment above about inferring metadata from the name

core/models/machine.py
```diff
@@ -283,7 +284,15 @@ def convert_glance_image(glance_image, provider_uuid, owner=None):
...
+
+    metadata_version_name = glance_image.get('version_name', default_version_name)
+    if not version_name:
+        version_name = metadata_version_name
+    if metadata_version_name != version_name:
+        logger.warn("Deprecation warning: Using version_name:%s but metadata defines 'application_version=%s'",
+                    version_name, metadata_version_name)
```
cdosborn: I would change this logic like so:
```
version_name = glance_image.get('version_name', default_version_name)
```
Make the metadata the source of truth, not something in the image name.

cdosborn: on further inspection what i'm advocating for already existed

core/models/machine.py
```diff
@@ -301,15 +310,16 @@ def convert_glance_image(glance_image, provider_uuid, owner=None):
         identity = None
     app_kwargs.update({
         'created_by': user,
-        'created_by_identity': identity
+        'created_by_identity': identity,
+        'uuid': glance_image.get('application_uuid')
```
cdosborn: Why are we setting the uuid from the glance metadata, what if doesn't exist in the metadata. Also what if an application exists with this uuid, it would result in an update to the existing app (so this outside image, would totally replace the existing one).

service/accounts/openstack_manager.py
```diff
@@ -520,10 +509,13 @@ def shared_images_for(self, image_id, status="approved"):
...
+        except glanceclient.exc.HTTPNotFound as exc:
+            # If image is not found, no projects should be added
+            pass
         except GlanceForbidden as exc:
-            if 'Only shared images have members' in exc.details:
-                return projects
-            raise
+            # Skip over exception if image visibility is public/private.
+            if 'Only shared images have members' not in exc.details:
+                raise
```
cdosborn: :+1:

service/tasks/monitoring.py
```diff
@@ -287,48 +289,58 @@ def distribute_image_membership(account_driver, cloud_machine, provider):
...
+        request_provider = has_machine_request.new_machine_provider
+        request_identifier = has_machine_request.new_machine.instance_source.identifier
+        if request_provider != db_machine.provider:
+            main_account_driver = get_account_driver(request_provider)
+            # Extend to include based on information in the machine request
+            request_shared_projects = main_account_driver.shared_images_for(request_identifier, None)
+            request_shared_projects = [p.name for p in request_shared_projects if p.name not in shared_project_names]
+            shared_project_names.extend(request_shared_projects)
+
+    shared_project_names.extend([p.name for p in shared_projects if p.name not in shared_project_names])
```
cdosborn: Why isn't this just:
```
names = map(lambda p: p.name, shared_projects)
```

Logic is pretty hard to follow

service/tasks/monitoring.py
```diff
@@ -287,48 +289,58 @@ def distribute_image_membership(account_driver, cloud_machine, provider):
...
     matching_users = list(parent_app.get_users_from_access_list().values_list('username', flat=True))
-    if len(matching_users) > 128:
-        # THIS IS A HACK: If there are LOTS of projects shared with the parent application, thats likely
-        # an accident. _ESPECIALLY_ if the list in the access list is smaller. so lets just ignore them.
-        # - SGregory
-        matching_users = []
-    shared_group_names.extend([user for user in matching_users if user not in shared_group_names])
-    return shared_group_names
+    shared_project_names.extend([user for user in matching_users if user not in shared_project_names])
+    return shared_project_names
```
cdosborn: Name everything separately and return a concatenation:
```
return <matching_names> + <shared_names> + <share_users>
```
Its not ideal to mutate a variable and move along. Its the difference between me being able to look at the signature of the method (i.e. the return stmt) and more or less know what it does, v.s. having to read the entire body.

service/tasks/monitoring.py
```diff
@@ -974,7 +995,8 @@ def _share_image(account_driver, cloud_machine, identity, members, dry_run=False
     cloud_machine_is_public = cloud_machine.is_public if hasattr(cloud_machine,'is_public') else cloud_machine.get('visibility','') == 'public'
     if cloud_machine_is_public == True:
         celery_logger.info("Making Machine %s private" % cloud_machine.id)
-        account_driver.image_manager.glance.images.update(cloud_machine.id, visibility='shared')
+        if not dry_run:
+            account_driver.image_manager.glance.images.update(cloud_machine.id, visibility='shared')
```
cdosborn: nested ifs are cancer:
```
if cloud_machine_is_public and not dry_run:
    do_thing
```

cdosborn: log stmts are also cancer

service/tasks/monitoring.py
```diff
@@ -287,48 +289,58 @@ def distribute_image_membership(account_driver, cloud_machine, provider):
...
+        shared_project_names.extend(
+                [name for name in access_list
+                 if name not in shared_project_names])
+        request_provider = has_machine_request.new_machine_provider
+        request_identifier = has_machine_request.new_machine.instance_source.identifier
+        if request_provider != db_machine.provider:
+            main_account_driver = get_account_driver(request_provider)
+            # Extend to include based on information in the machine request
+            request_shared_projects = main_account_driver.shared_images_for(request_identifier, None)
+            request_shared_projects = [p.name for p in request_shared_projects if p.name not in shared_project_names]
```
cdosborn: This and line 317 motivate the use of a Set.

service/tasks/monitoring.py
```diff
@@ -287,48 +289,58 @@ def distribute_image_membership(account_driver, cloud_machine, provider):
...
+    shared_project_names = [image_owner]
 
+    shared_projects = account_driver.shared_images_for(cloud_machine.id, None)
     # Extend to include based on projects already granted access to the image
-    shared_group_names.extend([p.name for p in shared_projects if p.name not in shared_group_names])
+    shared_project_names.extend([p.name for p in shared_projects if p.name not in shared_project_names])
 
-
-    #Extend to include new names found by application pattern_match
+    has_machine_request = MachineRequest.objects.filter(
```
cdosborn: there is a better place to lookup this information

core/models/machine.py
steve-gregory: Agreed. Will move away from this entirely.

core/models/application.py
```diff
@@ -50,6 +50,10 @@ def get_users_from_access_list(self):
         - Returns a list of Users who passed the test
         """
         from core.models import AtmosphereUser
+
+        if not self.access_list.count():
+            return AtmosphereUser.objects.none()
+
```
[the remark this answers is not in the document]
steve-gregory: Will provide test cases with expected results as part of this PR to clarify intentions, thanks for pointing this out!

core/models/machine.py
steve-gregory: For `glance images` that will be included in the Atmosphere image catalog,

If The `glance image` does not contain `application_uuid` in metadata,  it is the "Seed/First" version.
In this case, `create_application` will generate a UUID-hash with `glance_image.id`. Future lookups via monitoring can use `glance_image.id`  will always point to the created application.

Future updates to this image should come by the User issuing a `MachineRequest` to have a new image authored by admin identity.  After the MachineRequest is approved, the `processing` step of the imaging workflow is responsible for setting the metadata (`application_uuid, version_name, change_log, etc.`).

More updates on this process coming soon..

core/models/machine.py
steve-gregory: Method still exists, but is no longer in use within the codebase.
```
git grep _split_cloud_name
core/models/machine.py:    def _split_cloud_name(cls, machine_name):
```

core/models/application.py
```diff
@@ -50,9 +50,13 @@ def get_users_from_access_list(self):
...
         from core.models import AtmosphereUser
-        all_users = AtmosphereUser.objects.all()
+
+        all_users = AtmosphereUser.objects.none()
         for pattern_match in self.access_list.all():
-            all_users &= pattern_match.validate_users()
+            if pattern_match.allow_access:
+                all_users |= pattern_match.validate_users()
+            else:
+                all_users &= pattern_match.validate_users()
```
cdosborn: the &= case won't work, logic should be like: if allow_access all_users begins as the empty set, else all_users begins as everyone

core/models/application.py
```diff
@@ -354,6 +358,20 @@ class Meta:
         unique_together = ('application', 'group')
 
 
+def _get_owner_identity(provider, created_by=None):
+    """
+    This private method is intenteded to ensure that a valid identity exists
```
cdosborn: intentededededed :smile:

service/tasks/monitoring.py
```diff
@@ -287,48 +289,61 @@ def distribute_image_membership(account_driver, cloud_machine, provider):
...
     parent_app = db_machine.application_version.application
     matching_users = list(parent_app.get_users_from_access_list().values_list('username', flat=True))
-    if len(matching_users) > 128:
-        # THIS IS A HACK: If there are LOTS of projects shared with the parent application, thats likely
-        # an accident. _ESPECIALLY_ if the list in the access list is smaller. so lets just ignore them.
-        # - SGregory
-        matching_users = []
-    shared_group_names.extend([user for user in matching_users if user not in shared_group_names])
-    return shared_group_names
+    access_list_set = set(username for username in matching_users)
```
cdosborn: set(matching_users)

service/tasks/monitoring.py
```diff
@@ -287,48 +289,61 @@ def distribute_image_membership(account_driver, cloud_machine, provider):
...
+    - The owner of the application/Creator of the MachineRequest
+    - If using settings.REPLICATION_PROVIDER:
+      - include all those approved on the replication provider's copy of the image
+    """
+    #TODO: In a future update to 'imaging' we might image 'as the user' rather than 'as the admin user', in this case we should just use 'owner' metadata
 
+    image_owner = cloud_machine.get('application_owner','')
+    # NOTE: This assumes that the 'owner' (atmosphere user) == 'project_name' (Openstack)
+    # Always include the original application owner
+    owner_set = set([image_owner])
```
cdosborn: What if image owner is "", if we don't anticipate handling that case, image_owner = cloud_machine.get('application_owner'), which would produce a much better error message, than why the hell is the empty string a group

service/tasks/monitoring.py
```diff
@@ -287,48 +289,61 @@ def distribute_image_membership(account_driver, cloud_machine, provider):
...
 
-    #Extend to include new names found by application pattern_match
+    existing_members = account_driver.get_image_members(cloud_machine.id, None)
+    # Extend to include based on projects already granted access to the image
+    cloud_shared_set = set(p.name for p in existing_members)
```
cdosborn: if you want to be kool cat: `{ p.name for p in existing_members }` is equiv

core/models/application.py
steve-gregory: Submitted test cases, the logic works as expected for now. Future changes can come in a separate PR

steve-gregory:
Thanks to @cdosborn for the review(s)! :shipit:

PR #532, opened 2017-10-18
Problem: Machine validation is inflexible, site operators have different opinions on what "is valid"...

steve-gregory:
## Solution: Add MachineValidationPlugin support

- Rather than a single function to check if 'machine_is_valid', allow
  the atmosphere deployer to select the plugin that works for them.
- Currently there are three different machine validation plugins to
  chose from, separated into two files.
- atmosphere/plugins/machine_validation: AlwaysAllow, BasicValidation
  jetstream/plugins/machine_validation: BlacklistValidation
- New settings included: `MACHINE_VALIDATION_PLUGIN`

- [x] Documentation updated in [settings/local.py.j2](https://github.com/steve-gregory/atmosphere/blob/v29-plugin-for-machine-validation/atmosphere/settings/local.py.j2#L502-L508) to give context to the feature
- [x] [New default variables supported in Clank](https://github.com/cyverse/clank/pull/232)
- [x] New override variables committed to secrets repos (See clank PR for links..)


core/metrics/application.py
```diff
@@ -17,39 +17,39 @@
...
-            metrics = pickle.loads(pickled_object)
-        elif not read_only:
-            metrics = calculate_application_metrics(application, interval, day_limit, now_time=now_time)
-            pickled_object = pickle.dumps(metrics)
-            redis_cache.set(key, pickled_object)
-            redis_cache.expire(key, METRICS_CACHE_DURATION)
-    except:
-        logger.exception("Unexpected errror in application metrics")
-    return metrics
+# def get_application_metrics(application, now_time=None, read_only=False):
```
c-mart: [Delete Your Dead Code](https://late.am/post/2016/04/28/delete-your-dead-code.html)

core/metrics/application.py
```diff
@@ -103,92 +103,92 @@ def calculate_summarized_application_metrics(app):
 
 
 # Alternative calculation method, drilling down per-version rather than per-application.
-def calculate_detailed_application_metrics(application, interval=rrule.MONTHLY):
```
c-mart: Again, what does this have to do with "machine validation"? Changes for another PR, perhaps?

core/metrics/application.py
```diff
@@ -103,92 +103,92 @@ def calculate_summarized_application_metrics(app):
...
-    median_status_time = numpy.median(status_in_secs) if status_in_secs != 0 else 0
-    avg_status_time = numpy.mean(status_in_secs) if status_in_secs != 0 else 0
-    std_dev_status_time = numpy.std(status_in_secs) if status_in_secs != 0 else 0
-    status_stats = {
-        'average': avg_status_time,
-        'std_deviation': std_dev_status_time,
-        'median': median_status_time,
-    }
-    return status_stats
+# def calculate_detailed_application_metrics(application, interval=rrule.MONTHLY):
```
c-mart: Same as above

core/tasks.py
```diff
@@ -14,7 +14,7 @@
 from core.models.status_type import get_status_type
 from core.models.application import Application
 from core.query import only_current_apps
-from core.metrics.application import get_application_metrics
```
c-mart: More changes for a different PR?

jetstream/plugins/machine_validation.py
```diff
@@ -0,0 +1,35 @@
+from atmosphere.plugins.machine_validation import MachineValidationPlugin
+from django.conf import settings
+from threepio import logger
+
+
+class JetstreamValidation(MachineValidationPlugin):
+    """
+    Represents the current strategy being used by Jetstream
+
+    Default metadata_key is: `atmosphere_catalog`
```
c-mart: So `atmosphere_catalog="false"` would still result in the image being added to the catalog? How about we also require the metadata value to be truthy?

service/deploy.py
```diff
@@ -138,7 +138,7 @@ def deploy_unmount_volume(instance_ip, username, instance_id,
 
     #lsof returns 1 on success _and_ failure, so combination of 'rc' and 'stdout' is required
     if lsof_rc != 0 and lsof_stdout != "":
-        _raise_lsof_playbook_failure(lsof_rc, lsof_stdout)
```
c-mart: I don't understand how the changes in this file relate to the rest of the PR

service/mock.py
```diff
@@ -356,3 +359,81 @@ def attach_volume(self, *args, **kwargs):
...
+        PROJECT_SCHEMA = {'name': {'type': ['null', 'string'], 'description': 'name of the project/tenant', 'maxLength': 255}}
+        MockProject = warlock.model_factory(PROJECT_SCHEMA)
+        if not project_id:
+            return None
+        elif 'admin' in project_id.lower():
+            return MockProject(name='admin')
+        else:
+            return MockProject(name='UnknownUser')
+
+    def _generate_glance_image(self, **kwargs):
```
c-mart: Also unsure how this is related

atmosphere/plugins/machine_validation.py
```diff
@@ -0,0 +1,140 @@
...
+        elif owner_project.name != atmo_author_project_name:
+            logger.info(
+                "cloud machine %s - authored by Tenant %s, not the Atmosphere author: %s",
+                cloud_machine.id, owner_project.name, atmo_author_project_name)
+            return False
+        return True
+
+    def _is_kernel_or_ramdisk(self, cloud_machine):
+        cloud_machine_name = cloud_machine.name if cloud_machine.name else ""
+        if any(cloud_machine_name.startswith(prefix) for prefix in ['eri-', 'eki-', 'ari-', 'aki-']):
```
c-mart: -1 this sets a trap for users. My friend at the Edmonton Research Institute is about to create his "eri-genomics" image, and then become sad when it doesn't show up in Atmosphere.

atmosphere/plugins/machine_validation.py
```diff
@@ -0,0 +1,140 @@
...
+
+    def _is_snapshot(self, cloud_machine):
+        cloud_machine_name = cloud_machine.name if cloud_machine.name else ""
+        if cloud_machine_name.startswith("ChromoSnapShot"):
+            return True
+        if cloud_machine.get('image_type', 'image') == 'snapshot':
+            return True
+        return False
+
+    def _machine_in_same_domain(self, cloud_machine):
```
c-mart: What is the need for this check?

atmosphere/plugins/machine_validation.py
```diff
@@ -0,0 +1,140 @@
...
+    """
+    def machine_is_valid(self, cloud_machine):
+        """
+        Given a cloud_machine (glance image)
+
+        Return True if it passes the sanity checks.
+        """
+        return self._sanity_check_machine(cloud_machine)
+
+class CyverseValidation(MachineValidationPlugin):
```
c-mart: I disagree, see summary.

atmosphere/settings/__init__.py
```diff
@@ -465,12 +465,6 @@ def create_log_path(filename):
         "schedule": timedelta(minutes=60),
         "options": {"expires": 10 * 60, "time_limit": 2 * 60}
     },
-    "generate_metrics": {
```
c-mart: For a different PR?

atmosphere/plugins/machine_validation.py
steve-gregory: users with the Openstack 'admin' role see _all images, across all domains_

But that doesn't necessarily mean anyone on _the atmosphere-managed domain_ can launch the image. Best to skip it, in that case..

atmosphere/plugins/machine_validation.py
steve-gregory: Good point, i'll drop the name restrictions.

atmosphere/settings/__init__.py
steve-gregory: Consequence of removing code, in an attempt to pass test-cases before merging.

core/metrics/application.py
steve-gregory: I agree this code should be removed, in a separate PR.. It was commented out in an effort to "pass the tests".

jetstream/plugins/machine_validation.py
steve-gregory: Two of the things I wanted to discuss during the review:

1) What should we use as the default metadata_key
2) Do we want to treat metadata as boolean values, or simply require the existence of the key itself.

I'm fine either way..

service/deploy.py
steve-gregory: This was an error uncovered by running `prospector` during the `travis.ci tests` as part of PR building. It was fixed in an effort to make the PR review-ready.

service/mock.py
steve-gregory: The MockAccountDriver allows me to [write test cases](https://github.com/cyverse/atmosphere/pull/532/files#diff-968ed44e82322a9aca620da73c70a50fR19) that test images using the same logic as `monitor_machines_for` to ensure we get the desired results.

jetstream/plugins/machine_validation.py
steve-gregory: Follow-up after taking suggestions:
1. For "default-whitelist": `atmo_image_exclude`  and for "default-blacklist": `atmo_image_include`
2. Truthy! support `true/false` and `yes/no`

atmosphere/settings/__init__.py
cdosborn: @steve-gregory really the issue is that the base branch (i.e. someone else) merged code that was not passing tests, so every new pr against atmo will be breaking bec of unrelated code.

I'd submit pr to fix base branch tests, and then rebase this branch on top of the fixed base branch, so your pr is focused.

steve-gregory: Roger that. will do that today and look for a quick review-and-:shipit: to rebase

atmosphere/plugins/machine_validation.py
c-mart: roger!

jetstream/plugins/machine_validation.py
c-mart: sounds good @steve-gregory!

c-mart:
> Is the alternative to make the fixes, create a new PR for review, wait for merge, rebase accordingly?

Yep

> Keep the default, it will pick up all the images *you'd expect the Atmosphere image catalog to show*

Who would have which expectation exactly?

> I don't like the idea of isolating new site operators of Atmosphere (1), requiring that they change any/all of the images on their cloud to specific metadata.

Let's think in terms of policy vs. mechanism. Irrespective of how it is implemented, I think that the default *policy* for atmo(0,1,2) should be to only show images in the public catalog which have been 1. created from Atmosphere's imaging process, or 2. approved/selected by the operator. I believe that any cloud that the operator will do any testing/troubleshooting on (i.e. any cloud at all) will have a bunch of junk images that aren't fit for public display to atmo(1) users, and that expecting the operator to filter those out (as a default behavior) is either a bad experience for the operator (if filtering is attempted) or a bad experience for users (if operator doesn't bother).

It appears that you don't like the mechanism that Jetstream has proposed to solve this problem ("requiring that they change any/all of the images on their cloud to specific metadata" / "forced requirements on a cloud provider"). What about a different mechanism, perhaps like this?

- Atmo(2) periodically imports all images from OpenStack, but the public image catalog only shows images that have `catalog=True` property in atmo(2) DB, others are hidden
- Images created using the atmosphere(1) imaging process have `catalog=True` set on creation (so they will all appear in catalog)
- Operators can go into Django admin (or maybe Django shell / Tropo admin section) and set `catalog=True` for any other images they also wish to display in the catalog (e.g. those imported from cloud outside of image creation process).

steve-gregory:
> Who would have which expectation exactly?

Expectation: No snapshots should be shown, no kernel/ramdisk should be shown. You can't launch it and it doesn't fit the model.

> I don't like the idea of isolating new site operators of Atmosphere (1), requiring that they change any/all of the images on their cloud to specific metadata.

I still stand by this. We don't _need_ to force images out by metadata. There is another way...

> Let's think in terms of policy vs. mechanism. Irrespective of how it is implemented, I think that the default policy for atmo(0,1,2) should be to only show images in the public catalog which have been 1. created from Atmosphere's imaging process, or 2. approved/selected by the operator. 

Perfect! That is one strategy of validating machines for atmosphere! 

> I believe that any cloud ... will have a bunch of junk images that aren't fit for public display to atmo(1) users, and that expecting the operator to filter those out (as a default behavior) is either a bad experience for the operator (if filtering is attempted) or a bad experience for users (if operator doesn't bother).

One might argue that if a site operator (Who might not be a cloud admin) trying out Atmosphere, includes their account driver and provider, creates an identity to find they have instance(s), volume(s) and zero images they (and any potential users) are in an equally confusing experience.

If I added an Openstack Cloud to Atmosphere, _my expectation_ would be that all the images in `glance` (that my user can launch) are visible on the image list. 

With the addition of this strategy, and the two included implementations, we can help users and site operators get whatever it is that _they_ expect.

> What about a different mechanism, perhaps like this? Atmo(2) periodically imports all images from OpenStack, but the public image catalog only shows images that have catalog=True property in atmo(2) DB, others are hidden
> Images created using the atmosphere(1) imaging process have catalog=True set on creation (so they will all appear in catalog)
> Operators can go into Django admin (or maybe Django shell / Tropo admin section) and set catalog=True for any other images they also wish to display in the catalog (e.g. those imported from cloud outside of image creation process).

I'm against the idea of adding yet-another-attribute to the Instance DB, because we already  provide a method for the site operator to curate what is and what isn't shown in the catalog (end-dating).

We also provide a method (versions) to group together images that have a common purpose, abstracting away the cloud provider. (Admittedly, these features could use some love to make the process more user-friendly). Some operators may want their users to see every public image as an application (utilizing a flat structure). Some may curate all their individual machines into similar applications and versions (utilizing a nested/deep structure).

c-mart:
> One might argue that if a site operator (Who might not be a cloud admin) trying out Atmosphere, includes their account driver and provider, creates an identity to find they have instance(s), volume(s) and zero images they (and any potential users) are in an equally confusing experience.

Or, the install docs could just say "now, go to this page and check the boxes (or set a start date) for images you'd like displayed in the catalog". 

> _my expectation_ would be that all the images in glance (that my user can launch) are visible on the image list.

OK, so we disagree on the default policy, not just the mechanism. Maybe we have different ideas about the purpose of the image catalog. If a primary goal(??) of Atmosphere(1) is to make the cloud computing experience more user-friendly, then I think most deployers would want browsing/searching the image catalog to be a good experience rather than a confusing one, so I think most deployers would want to show things that are likely to be of use, but hide testing/troubleshooting/etc images. Maybe you don't see the catalog in this way?

> I'm against the idea of adding yet-another-attribute to the Instance DB, because we already provide a method for the site operator to curate what is and what isn't shown in the catalog (end-dating).

OK, then images could be imported with no start date set, and an operator could make an image visible by setting it to the current date/time... but again we're talking about the mechanism when we disagree on policy.

---

I'm advocating to change the default policy because the current default behavior pollutes the image catalog and degrades the user experience for all current atmosphere(1) deployments, but you're right, as long as we support a way to curate/whitelist, the default is less important.

In the interest of the Jetstream team getting this fixed soon, I suggest the following compromise: since these two strategies are not necessarily specific to CyVerse and Jetstream, rename the `CyverseValidation` and `JetstreamValidation` classes to `BlacklistValidation` and `WhitelistValidation`, moving the latter from `jetstream/plugins/machine_validation.py` to `atmosphere/plugins/machine_validation.py`. What do you think?

Related suggestions:
- Document the choices and how to use them in atmosphere-guides (perhaps in [installation](https://cyverse.github.io/atmosphere-guides/install_guide.html) or [Connecting a Cloud provider to Atmosphere](https://cyverse.github.io/atmosphere-guides/connecting_cloud_provider.html))
- It appears that `test_monitoring_with_blacklist_validation` is actually using "jetstream_validation" (a whitelist, not a blacklist), perhaps needs a fix to either the logic or the name
- Does Jetstream want images created using the Atmosphere(1) imaging process to be whitelisted (displayed) by default (i.e. whitelist metadata set by default)? Because right now I'm unsure if it will be.

steve-gregory:
In the interest of time and getting this fix out ASAP, I'll just say "Yes".

Post-review ToDo to :shipit: :
- [ ] Update atmosphere guides
- [ ] Rename machine_validation plugins, change file location(s) accordingly.
- [ ] Ensure that images are whitelisted by default through atmosphere imaging process.

PR #535, opened 2017-10-21
Problem: ansible fails when script text includes special characters.

steve-gregory:
## Solution

Remove special characters before sending script_text to ansible.
Provide a test case to prove that the solution works.


steve-gregory:
Seeking review -- Will self-:shipit: this bugfix in 24hr if there are no objections

steve-gregory:
No objections. Self-:shipit:

PR #539, opened 2017-11-02
Problem: Instances with ephemeral storage mounted to /home cannot be shelved

steve-gregory:
## Solution
Create a metadata 'escape hatch' for instances with ephemeral drive mounted to /home. 
Treat these instances differently by ensuring any call to 'Shelve' will actually suspend to avoid losing data.


steve-gregory:
self-:shipit:

PR #540, opened 2017-11-03
Problem: Including providers in fixtures can cause a "bounce-back" of end-dated values

steve-gregory:
## Solution
Remove the default providers from fixtures. Only the `ProviderType` and `PlatformType` are necessary.


steve-gregory:
self-:shipit:

PR #544, opened 2017-11-15
Option 2: Remove subspace in favor of ForkedAnsible

steve-gregory:
In this world, we would use [a forked ansible with newly inserted code.](https://github.com/cyverse/ansible/pull/1)

### DIFF
--- a/service/deploy.py
+++ b/service/deploy.py
@@ -1,7 +1,9 @@
 """
 Deploy methods for Atmosphere
 """
+import operator
 import os
+import json
 import re
 
 from django.template.loader import render_to_string
@@ -9,7 +11,6 @@ from django.utils.text import slugify
 from django.utils.timezone import datetime
 
 
-import subspace
 
 from threepio import logger, deploy_logger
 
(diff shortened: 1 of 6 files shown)


c-mart:
Spoke with @steve-gregory about this plan last week, and I'm on board with it:
- For now, use [forked Ansible](https://github.com/cyverse/ansible/pull/1) with our changes to provide additional functionality. Stop using custom Subspace middleware.
- Make a good faith effort to contribute these enhancements upstream to Ansible

service/deploy.py
```diff
@@ -339,75 +357,77 @@ def user_deploy_install(instance_ip, username, instance_id, install_action, inst
...
+    directories.sort(cmp=operator.lt)
+    for d in directories:
+        a_dir = d[0]
+        files_in_dir = d[2]
+        files_in_dir.sort()
+        if os.path.isdir(a_dir):
+            for f in files_in_dir:
+                if os.path.splitext(f)[1] == ".yml":
+                    files.append(os.path.join(a_dir, f))
+    return files
```
cdosborn: A lot of this code could be removed if we leveraged the fact that you can create playbooks which contain other playbooks. We shouldn't be sorting by name. Could create a playbook instance_deploy.yml that includes the playbooks it needs.

steve-gregory: Interesting idea @cdosborn - Let's get this in _first_ and then make that change separately with a corresponding PR in `atmosphere-ansible`?

steve-gregory:
I'm going to make an effort to get this merged in on Friday, but i'd like to work with @cdosborn or one of the other back-end developers to ensure that `cyverse-ansible` can be signed/packaged without issue.

cdosborn:
Closed in favor of https://github.com/cyverse/atmosphere/pull/631

PR #546, opened 2017-11-17
Fixes to monitor_machines and validation

steve-gregory:
## Highlights
- Legacy clouds need to call 'list images' twice and append  info to the v2 api.
- Skip machines if their status is 'queued' or 'saving'


steve-gregory:
self-:shipit:

PR #548, opened 2017-11-21
Problem: Allocations should renew on first day of calendar month

julianpistorius:
Solution: Refactor rules parser and add a `on_calendar_day` strategy

### DIFF
--- a/cyverse_allocation/cyverse_rules_engine_setup.py
+++ b/cyverse_allocation/cyverse_rules_engine_setup.py
@@ -1,5 +1,5 @@
 from business_rules.actions import BaseActions, rule_action
-from business_rules.fields import FIELD_NUMERIC
+from business_rules.fields import FIELD_NUMERIC, FIELD_TEXT
 from business_rules.variables import BaseVariables, boolean_rule_variable, numeric_rule_variable, string_rule_variable
 
 from core.models.allocation_source import AllocationSource,AllocationSourceSnapshot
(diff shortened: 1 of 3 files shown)


julianpistorius:
Notes after discussion with @steve-gregory :

There is a pernicious problem: What happens when the scheduled job for renewing allocations does not run on the 1st day of the month?

Possible solutions:

1. Logic that takes into account missed days due to the scheduled task not running (or not finishing) on the calendar day in question.

2. A 'hit the button' strategy that an administrator can trigger manually. I think this is already possible with existing scripts, and I will confirm and document that process.

There is another problem: What happens when the allocation is to renew on the 31, 30, or 29th, but the current month does not have that many days? In this case a 'last_day_of_month' strategy is probably the simplest solution.

julianpistorius:
Checked and yes, there is a script that does this:

`scripts/force_renew_allocation_sources.py`

julianpistorius:
@lenards The Travis failure was fixed in a separate PR #549

cyverse_allocation/cyverse_rules_engine_setup.py
```diff
@@ -80,79 +77,144 @@ def cannot_renew_allocation_source(self):
...
+                    "compute_allowed": compute_allowed
+                }
+        }
+    )
 
     return actions
 
 
 # RENEWAL STRATEGY CONFIGURATION
 renewal_strategies = {
```
lenards: it might be helpful to include some treatment of "key is `strategy_name` for the `renewal_startegies` might help to convey the data-definition. 

Not necessary - but for an outlander like me, it made that "jump" and I think I'm right. Just would be cool to _know_, :100:.

lenards:
:shipit:

PR #559, opened 2017-12-08
Fix unbound local variable username

cdosborn:
## Description

username was throwing unbound because request.session.get(...) was throwing an
attribute error. The attribute error was thrown because the request object
doesn't have a session attribute in a test environment, because the test
environment doesn't include SessionMiddleware.

So its an error to assume that username would be given a value, that was fixed
by username = None, and its equally bad to catch all exceptions, that was
fixed


julianpistorius:
Looks good! 🎉

variables.ini.dist
```diff
@@ -95,7 +95,7 @@ SUPPORT_EMAIL_SIGNATURE = # CyVerse Atmosphere Team
 SUPPORT_LINKS = {} # {'getting_started': "https://link_to.wiki.org/Using+Instances", 'new_provider': "https://link_to.wiki.org/Changing+Providers", "faq": "https://link_to.wiki.org/Cloud_FAQ"}
 REPLICATION_PROVIDER_LOCATION = # iPlant Cloud - Tucson
 MONTHLY_RESET_PROVIDER_LOCATIONS = [] # ['iPlant Cloud - Tucson']
-USER_EMAIL_LOOKUP_METHOD = # ldap_get_email_info
+USER_EMAIL_LOOKUP_METHOD = "django_get_email_info" # If ldap, ldap_get_email_info
```
cdosborn: When we run tests we generate variables.ini from this dist file. It's really just a mistake to have all these variables be empty.

PR #569, opened 2018-01-05
WIP Problem: Special allocation quota not set on first login

julianpistorius:
Due to two separate plugins which don't play well together. My tests
for the special allocation plugin does not try the combination with this
other plugin. `DefaultQuotaPluginManager` is never triggered when
`AccountCreationPluginManager` is enabled.

Solution: Temporary workaround - a script run in cron to reset the quota

- [x] If necessary, include a snippet in CHANGELOG.md

### DIFF
--- a/scripts/enforce_special_allocation_quota.py
+++ b/scripts/enforce_special_allocation_quota.py
@@ -1,0 +1,55 @@
+#!/usr/bin/env python
+import argparse
+
+import django
+
+django.setup()
+
+from core.models import Identity
+
+from service.quota import set_provider_quota
+
+
+def main():
+    """
+    Return a list of ALL users on a provider, their CURRENT allocation totals,
+    and # of instances used.
+    """
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--dry-run', action='store_true', help='Print output rather than perform operation')
+    parser.set_defaults(dry_run=False)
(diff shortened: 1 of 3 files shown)


scripts/enforce_special_allocation_quota.py
```diff
@@ -0,0 +1,55 @@
...
+     LEFT JOIN public.user_allocation_source ON atmosphere_user.id = user_allocation_source.user_id
+     LEFT JOIN public.allocation_source ON user_allocation_source.allocation_source_id = allocation_source.id
+   GROUP BY atmosphere_user.id
+   HAVING array_agg(allocation_source.name) = '{TG-ASC160018}'
+   ORDER BY username) AS users
+  LEFT OUTER JOIN public.identity ON identity.created_by_id = users.id
+  LEFT OUTER JOIN public.quota ON identity.quota_id = quota.id
+WHERE
+  identity.quota_id <> 33
+  AND identity.quota_id IS NOT NULL;'''
```
cdosborn: AND condition here doesn't make sense, if id is 33 then its guaranteed to be not null

cdosborn: Oh i'm wrong here. Misread <>

julianpistorius:
After talking to Connor I'm not going to commit this temporary script.

PR #565, opened 2017-12-18
Enforcement override

julianpistorius:
Problem: Need admin override for overage enforcement

Required behavior:
- Ignore enforcement/always allow: Allow allocations to "run" regardless
of SUs remaining in the allocation
- Always block: Prevent allocations or accounts (as appropriate) from
"running" regardless of SUs remaining in the allocation

Solution:
- Added a `get_enforcement_override` method on the allocation
source plugins
- The `monitor_allocation_sources` task checks the return value of this
method for every user + allocation source combination. If the value is:
  - `NEVER_ENFORCE`: Enforcement is always skipped
  - `ALWAYS_ENFORCE`: Enforcement is always triggered
  - `NO_OVERRIDE`: Enforcement depends on the `compute_used` (default)

For the initial version of this feature both the CyVerse & Jetstream
plugins will check the Django settings for two variables:

- `ALLOCATION_OVERRIDES_NEVER_ENFORCE`
- `ALLOCATION_OVERRIDES_ALWAYS_ENFORCE`

These should be lists with allocation source names.

Note: Global setting `ENFORCING` overrides any other behavior.
So if an allocation source has an override of `ALWAYS_ENFORCE`, but
`ENFORCING = False`, then no enforcement action will occur.

Proposed future enhancements:

- An admin panel where staff can configure this feature

## Checklist before merging Pull Requests
- [x] New test(s) included to reproduce the bug/verify the feature
- ~[ ] Documentation created/updated at [Example link to documentation](https://example.test/doc#new_section) to give context to the feature~
- ~[ ] If creating/modifying DB models which will contain secrets or sensitive information, PR to [clank](https://github.com/cyverse/clank) updating sanitation queries in `roles/sanitary-sql-access/templates/sanitize-dump.sh.j2`~
- [x] Reviewed and approved by at least one other contributor.
- [x] If necessary, include a snippet in CHANGELOG.md
- ~[ ] New variables supported in Clank~
- ~[ ] New variables committed to secrets repos~

PR #578, opened 2018-02-20
Problem: Use enforcement override for instance actions

julianpistorius:
Based on previous work: https://github.com/cyverse/atmosphere/pull/565

- Allow instances to always be started/resumed/unshelved if they are
    associated with an allocation source which is set to `NEVER_ENFORCE`
- Prevent instances from being started/resumed/unshelved if they are
    associated with an allocation source which is set to `ALWAYS_ENFORCE`

Solution:
- Run `check_allocation` in `service.instance.run_instance_action` if
    the action is in `('start', 'resume', 'unshelve')`

### DIFF
--- a/service/instance.py
+++ b/service/instance.py
@@ -19,6 +19,8 @@ from rtwo.exceptions import LibcloudBadResponseError
 from rtwo.driver import OSDriver
 from rtwo.drivers.openstack_user import UserManager
 from rtwo.drivers.common import _token_to_keystone_scoped_project
+
+from core.plugins import AllocationSourcePluginManager, EnforcementOverrideChoice
 from service.driver import AtmosphereNetworkManager
 from service.mock import AtmosphereMockDriver, AtmosphereMockNetworkManager
 
@@ -32,7 +34,7 @@ from libcloud.common.exceptions import BaseHTTPError  # Move into rtwo.exception
 
 from core.query import only_current
 from core.models.instance_source import InstanceSource
-from core.models import AtmosphereUser
+from core.models import AtmosphereUser, InstanceAllocationSourceSnapshot
 from core.models.ssh_key import get_user_ssh_keys
 from core.models.application import Application
 from core.models.identity import Identity as CoreIdentity
@@ -1315,12 +1317,27 @@ def _test_for_licensing(esh_machine, identity):
 
 
 def check_allocation(username, allocation_source):
+    logger.debug('check_allocation - username: %s', username)
+    logger.debug('check_allocation - allocation_source: %s', allocation_source)
+    if not settings.ENFORCING:
+        logger.debug('check_allocation - settings.ENFORCING is False - just returning')
+        return
     user = AtmosphereUser.objects.filter(username=username).first()
     if not user:
         raise Exception("Username %s does not exist" % username)
+
+    enforcement_override_choice = AllocationSourcePluginManager.get_enforcement_override(user,
+                                                                                         allocation_source)
+    logger.debug('check_allocation - enforcement_override_choice: %s', enforcement_override_choice)
+    if enforcement_override_choice == EnforcementOverrideChoice.NEVER_ENFORCE:
+        return
+    elif enforcement_override_choice == EnforcementOverrideChoice.ALWAYS_ENFORCE:
+        # Using a dummy value of -1, should throw a separate error
+        raise OverAllocationError(allocation_source.name, -1)
+
     compute_remaining = allocation_source.time_remaining(user)
     over_allocation = compute_remaining < 0
-    if over_allocation and settings.ENFORCING:
+    if over_allocation:
         raise OverAllocationError(allocation_source.name, abs(compute_remaining))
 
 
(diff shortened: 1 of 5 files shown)


service/instance.py
```diff
@@ -1315,12 +1317,31 @@ def _test_for_licensing(esh_machine, identity):
...
     if not user:
         raise Exception("Username %s does not exist" % username)
     compute_remaining = allocation_source.time_remaining(user)
     over_allocation = compute_remaining < 0
-    if over_allocation and settings.ENFORCING:
+
+    enforcement_override_choice = AllocationSourcePluginManager.get_enforcement_override(user,
+                                                                                         allocation_source)
+    logger.debug('check_allocation - enforcement_override_choice: %s', enforcement_override_choice)
+    if over_allocation and enforcement_override_choice == EnforcementOverrideChoice.NEVER_ENFORCE:
```
cdosborn: You can simplify the logic here for a readability win. The user's compute remaining doesn't actually matter (and could be computed after this if check).
```
    if enforcement_override_choice == EnforcementOverrideChoice.NEVER_ENFORCE:
        return
    elif enforcement_override_choice == EnforcementOverrideChoice.ALWAYS_ENFORCE:
        # Using a dummy value of -1, should throw a separate error
        raise OverAllocationError(allocation_source.name, -1)
```

service/instance.py
```diff
@@ -2035,6 +2056,10 @@ def run_instance_action(user, identity, instance_id, action_type, action_params)
         action_type = 'suspend'
 
     logger.info("User %s has initiated instance action %s to be executed on Instance %s" % (user, action_type, instance_id))
+    if action_type in ('start', 'resume', 'unshelve'):
+        logger.info('Going to check the allocation of the instance due to the action type: %s', action_type)
```
cdosborn: See comment above w.r.t to logging. This may be useful while you're developing the feature, but would argue shouldn't be part of the codebase.

cdosborn:
Before this pr goes in, it needs to be rebased onto latest v30. When you do that you should see that tests are failing. I'm seeing about 7 scenario failures.

cdosborn:
Okay when I rebased it on top of master, my travis succeeded so something is wrong with my dev environment.

service/instance.py
julianpistorius: I did consider using simpler logic like you have, but decided against it. I chose to use this more explicit and verbose logic with the matching logging statements to help some poor future programmer to figure out how this feature is supposed to work.

service/instance.py
```diff
@@ -1315,12 +1317,31 @@ def _test_for_licensing(esh_machine, identity):
...
+    if over_allocation and enforcement_override_choice == EnforcementOverrideChoice.NEVER_ENFORCE:
+        logger.debug('Allocation source is over allocation, but %s + user %s has an override of %s, '
+                     'therefore not enforcing',
+                     allocation_source, user, enforcement_override_choice)
+        return
+    if not over_allocation and enforcement_override_choice == EnforcementOverrideChoice.ALWAYS_ENFORCE:
+        logger.debug('Allocation source is not over allocation, but %s + user %s has an override of %s, '
+                     'therefore enforcing',
+                     allocation_source, user, enforcement_override_choice)
+        raise OverAllocationError(allocation_source.name, -compute_remaining)
```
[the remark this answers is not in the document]
julianpistorius: Agreed. I started implementing this 'sensible exception', but then I looked into what would be involved and saw how much code I would have to touch, I decided against going down this route for this particular PR due to the short time we have to test it.

Your suggestion is definitely the right way to do it, but I'd like do do it in a follow-up improvement PR.

service/instance.py
julianpistorius: I prefer to use logging statements like this to help debugging issues. It shouldn't be a problem when you have a good log aggregator with flexible filtering rules.

service/instance.py
```diff
@@ -1315,12 +1317,31 @@ def _test_for_licensing(esh_machine, identity):
 
 
 def check_allocation(username, allocation_source):
+    logger.debug('check_allocation - username: %s', username)
```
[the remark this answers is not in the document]
julianpistorius: Fair enough. I've struggled trying to figure out production issues too many times with not enough clues. That's why err on the side of many log statements, and using tools to like log aggregation to manage the cognitive overhead.

julianpistorius:
@cdosborn Thanks for the thoughtful & thorough review.

julianpistorius:
@cdosborn I've rebased over latest v30.

service/instance.py
cdosborn: The current logic was unclear to me, and I think there's more to it than a diff of opinion. 

This code has 1 of 2 behaviors and the fresh reader is going to try and understand how ALWAYS/NEVER work. Either ALWAYS/NEVER are absolute, or they only come into effect when the user is over. If they see that over allocation is part of the function, they should assume that ALWAYS/NEVER play a role when as is over_allocation. However, the actual behavior is that these checks don't take over_allocation into account at all.

julianpistorius: Hmm... Fair enough. I see your point. I'll modify it.

service/instance.py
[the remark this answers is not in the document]
cdosborn: When you log this often (or imagine in each sub routine) then when multiple threads/processes are executing code you get spaghetti in log files (or you have to invest in more complicated logging solution to handle this). 

I want to see tracebacks (exceptions/errors) in log files which give immediate context with files and line nos, no trail of messages i have to search. I also want to look at a function and see as little noise as possible. We can ignore this for now. But would like to reach consensus for future prs.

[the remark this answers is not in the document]
julianpistorius: I definitely sympathize. It is a delicate balance.

The noisy logs and context problem can be solved by tools like Graylog/ELK and something like Eliot: https://github.com/ScatterHQ/eliot

service/instance.py
julianpistorius: @cdosborn Change made.

cdosborn:
This fixes the broken test
https://github.com/cyverse/atmosphere/pull/580

PR #584, opened 2018-02-27
Remove unnecessary enforcing restrictions

cdosborn:
## Description

Problem: Testing allocation logic requires setting ENFORCING=true, which is dangerous
Solution: Remove ENFORCING from the equation

~~- [ ] New test(s) included to reproduce the bug/verify the feature~~
~~- [ ] Documentation created/updated at [Example link to documentation (https://example.test/doc#new_section) to give context to the feature~~
~~- [ ] If creating/modifying DB models which will contain secrets or sensitive information, PR to [clank](https://github.com/cyverse/clank) updating sanitation queries in `roles/sanitary-sql-access/templates/sanitize-dump.sh.j2`~~
~~- [ ] If necessary, include a snippet in CHANGELOG.md~~
~~- [ ] New variables supported in Clank~~
~~- [ ] New variables committed to secrets repos~~

### DIFF
--- a/api/tests/v2/test_instance_actions.py
+++ b/api/tests/v2/test_instance_actions.py
@@ -12,7 +12,10 @@ from api.tests.factories import (
     ApplicationVersionFactory, ProviderMachineFactory, IdentityFactory, ProviderFactory,
     SizeFactory, AllocationSourceFactory)
 from api.v2.views import InstanceViewSet
-from core.models import InstanceAllocationSourceSnapshot
+from core.models import (
+    InstanceAllocationSourceSnapshot,
+    UserAllocationSnapshot, AllocationSourceSnapshot,
+    UserAllocationSource)
 from service.driver import get_esh_driver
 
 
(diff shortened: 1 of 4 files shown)


cdosborn:
As a result of this patch, 2 tests are now running more code than they used to. I believe they are failing because a bug was uncovered here: https://github.com/cyverse/atmosphere/pull/585. I will test if that fix also fixes these failing tests.

cdosborn:
Here is the test to see if thats the case:
https://travis-ci.org/cdosborn/atmosphere/builds/347007615

cdosborn:
Okay tests are failing because they were testing with allocation sources that weren't created properly will fix those too SMH.

cdosborn:
Fixed.

cdosborn:
This pr includes https://github.com/cyverse/atmosphere/pull/585 (just a single commit f11b796427a49522b6718c1a121f2935f9c8575d), when that is merged, ill rebase this pr so its not included here.

PR #586, opened 2018-02-28
Guacamole web shell and desktop link generation update

calvinmclean:
## Description

Previous versions of Guacamole required that the authentication plugin removed the first two characters of the connection ID in order to add some sort of extra metadata the the ID. This was removed, so the auth plugin no longer removes the two characters, so neither should Atmosphere.

**Before this goes into production**, please coordinate with me to make sure Guacamole server has the same changes because they have to go at the same time.


cdosborn:
Confirmed that this works!

PR #592, opened 2018-03-03
Fix image scripts not being included in user deploy

cdosborn:
## Description

**Problem:**
Image scripts are never deployed to an instance

**Solution:**
Deploy image scripts alongside instance scripts

In this commit acb6729544449b352401a307c8300b1ab2cd6d8b, we removed the task which ran image scripts against an instance. 


cdosborn:
My original patch was so small it seemed like it would just work (and I did test it locally). It turned out not to work. Because you cannot filter over a queryset that has been evaluated. (you can, but you get nonsense results). Anywho to the power of writing tests for even the smallest changes.

service/tests/test_deploy.py
```diff
@@ -0,0 +1,36 @@
+from django.test import TestCase
+import mock
+
+from api.tests.factories import BootScriptRawTextFactory, InstanceFactory, UserFactory
+
+
+class UserDeployTests(TestCase):
```
julianpistorius: 💯

api/tests/v2/test_image_metrics.py
```diff
@@ -172,6 +172,7 @@ def test_user_sees_no_statistics(self):
         response = client.get(url)
         self.assertEquals(response.status_code, 404)
 
+    @skip("Test is non deterministic and yields different results based on its redis cache")
```
julianpistorius: Ouch.

PR #596, opened 2018-03-14
Change maintenance command to be non-interactive

cdosborn:
## Description
Change ./manage.py maintenance sub-command to be easier to test and add features

I was making changes to `atmosphere.version.git_branch`, and found that this command would be affected. I wanted to write some tests to ensure that the command would still work. However, the previous command would not allow you to pass some arguments and not others. If you omitted any, it would go into an interactive mode. I couldn't test whether the defaults were working, because of this mode. The command has been pretty useful during deployments, so these changes just make the tool  more command-line friendly.

## Other changes:
- Defaults are used if not overriden
- Interactive mode is removed and replaced with --dry-run
- The start_date format accepts timezones now.
- The --start-now flag was removed, this is the default if no --start-date is specified.
- The show command will print a current maintenance if one exists

- [x] Documentation created/updated (include links)
  - https://github.com/cyverse/atmosphere-guides/pull/39

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 ### Changed
   - Update license to 2018
   - Updated CHANGELOG.md to use the recommendations from [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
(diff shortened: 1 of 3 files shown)


cdosborn:
It's a shame coverage is reporting 0%.

PR #599, opened 2018-03-29
Fix unshelved instance failing to get fixed ip

cdosborn:
### Problem
On unshelve instances would sometimes fail to get a fixed ip

### Solution
Rather than update ip on current port, delete all ports and create a new one

I changed the name of add_fixed_ip, because we are taking a quite different approach here. We delete all an instance's ports, and then instruct nova to attach the instance to a subnet. Nova takes care of creating a port.

This fixes an outstanding issue in nova[0]. Prior to this change we kept ports around and then removed/added fixed ips. However, when an instance and its port are created, they each have an availability zone(AZ), when we unshelve an instance its AZ may change. If it does change, openstack fails to update the device owner AZ of the port and attaching a fixed ip fails because no ports exist in the AZ.

https://bugs.launchpad.net/nova/+bug/1759924 [0]

### DIFF
--- a/service/instance.py
+++ b/service/instance.py
@@ -470,7 +470,7 @@ def restore_ip_chain(esh_driver, esh_instance, deploy=True,
     start with: task.apply_async()
     """
     from service.tasks.driver import (
-        wait_for_instance, add_fixed_ip, add_floating_ip,
+        wait_for_instance, restore_port, add_floating_ip,
         deploy_init_to, update_metadata
     )
     init_task = wait_for_instance.s(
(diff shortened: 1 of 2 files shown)


c-mart:
I like the approach here [iff](https://en.wikipedia.org/wiki/If_and_only_if) we're continuing to remove fixed IP addresses from instances when we shelve (and suspend, etc.) them.

cdosborn:
We are still removing the fixed ips. However, instances affected by the nova bug will actually fail to remove the fixed ips, and restore_port compensates for that.

cdosborn:
My base pr was off, there is only a single commit.

PR #600, opened 2018-04-03
Bring in the latest from v31

cdosborn:
## Description

This pr https://github.com/cyverse/atmosphere/pull/599 was brought into v31, and thus v31 needs to now be brought into master


cdosborn:
I'm going to make sure this builds, then ill push it to master, so an unnecessary merge isn't made.

PR #604, opened 2018-04-13
Make another attempt to fix adding fixed ips

cdosborn:
## Description
Try different approach to working around openstack port availability zone issue

### Problem
After upgrading Marana to Pike we found that ports which were attached post boot, would never successfully attach. This meant that every time an instance would boot (start/unshelve) minus initial launch, it would not be reachable. A workaround was discovered to enable allow-hotplug for the given interface in cloud-init, but it was deemed more complicated than simply preserving the ports.

We were deleting/recreating the ports in the first place, because of a bug which affected j7m https://bugs.launchpad.net/nova/+bug/1759924.

### Solution

Rather than add/delete ports, add/remove fixed ips, but explicitly check if we're dealing with the availability zone issue and then try deleting/recreating the port. It is unknown why j7m is able to attach ports post boot, and we cannot.

I tested the following scenarios. Start/stop, shelve/unshelve on marana and indiana. (and in particular the shelve scenario where the AZ was incorrect)

### DIFF
--- a/service/tasks/driver.py
+++ b/service/tasks/driver.py
@@ -179,23 +179,44 @@ def add_fixed_ip(
         identity,
         instance_id,
         core_identity_uuid=None):
-    from service import instance as instance_service
+    from service.instance import _to_network_driver, _get_network_id
     try:
         celery_logger.debug("add_fixed_ip task started at %s." % datetime.now())
+        core_identity = Identity.objects.get(uuid=core_identity_uuid)
+        network_driver = _to_network_driver(core_identity)
         driver = get_driver(driverCls, provider, identity)
         instance = driver.get_instance(instance_id)
         if not instance:
             celery_logger.debug("Instance has been teminated: %s." % instance_id)
             return None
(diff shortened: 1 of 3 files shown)


service/tasks/driver.py
```diff
@@ -168,35 +168,61 @@ def _is_instance_ready(instance, status_query,
...
+            # If the port and instance are in different zones, delete the old
+            # port and attach a new one, this is only occurs in narrow
+            # scenarios documented in the following ticket:
+            # https://bugs.launchpad.net/nova/+bug/1759924
+            network_driver.delete_port(port)
+            driver._connection.ex_attach_interface(
+                instance.id, network_id=network_id)
+
+        elif not instance._node.private_ips:
+            # Only add fixed ip, if the instance doesn't already have on
```
c-mart: typo, "doesn't already have on**e**"

cdosborn: fixed

PR #607, opened 2018-04-17
Resolve pip-10 travis build failure

cdosborn:


cdosborn:
The build passed, going to bring this in to unblock other work depending on travis build.

PR #608, opened 2018-04-17
Preserve private networks and private/fixed IPs for inactive instances

c-mart:
## Description

This PR removes the code that removes private network resources from inactive (suspended/stopped/shelved instances). This effectively leaves the private subnet and fixed IP intact. We are still removing floating (public) IP addresses.

## Background

Currently, Atmosphere(2) removes fixed IP addresses (and possibly private networks) from instances when they are stopped, suspended, or shelved, and also periodically on all 'inactive' instances. This was originally done for at least the following reasons:

> * support for isolated tenant subnets, where each tenant gets their own subnet of /24 (though in hindsight /24 is probably too large)
> * at the time, neutron overlapping ips didn't work for us as expected or there was some gotcha, which I forget
> * because we assumed that subnet space would be finite and to plan for future growth and larger scale, then atmosphere would reclaim unused subnets (esp given that some users will use atmosphere infrequently or only once e.g. a workshop environment)
> * in order to be able to claim unused subnets, we needed to ensure that private ips were reclaimed cleanly, which meant a tear down of the network and port at the vm level

It has also been mentioned that the OpenStack Neutron service (in OpenStack Havana, or earlier) had capacity issues with many concurrent private networks.

I have seen us chase several bugs and spend many engineering hours dealing with consequences of (and workarounds for) the decision to remove private network resources from inactive instances, most recently during the v32 deployment, #599 and #604. I believe that we don't need to be doing this. At best, I think it solves a problem that we don't have and are unlikely to have in the near future.

I suggest the following:
- There are nearly 18 million IP addresses in RFC 1918 (private) space; that's 70,000 /24 subnets (which is many times the total number of users in any Atmosphere(2) deployment), or e.g. 280,000 /22 subnets
- If we do approach 70,000 active users someday (a great problem to have!) then we could easily switch to using smaller private subnets, or justify the engineering effort to deal with the consequences of removing private networks from inactive instances, using overlapping private networks, etc. In other words, I think this decision is easy to change later if needed.
- I suspect that our modern, load-balanced, highly available Neutron control plane can handle many concurrent private networks/subnets.
- I think that reclaiming private networks/subnets from users that have zero instances (active or inactive) is a safe action that is unlikely to cause bugs, though that is not implemented here

## Todo
- [x] Deploy changes to atmobeta
- [x] Stop and start instances
- [x] Suspend and resume instances
- [x] shelve-offload and unshelve instances
- [x] instances that are suspended/shelved/stopped now display the private IP address in Troposphere. This is likely to confuse users.
- ~[ ] When resuming instance, this is logged: `2018-04-17 13:59:02,523 atmosphere-WARNING [/opt/dev/atmosphere/service/networking.py 191] Unable to create new_cidr for subnet for user: test-cmart (CIDR already used)`~ Pre-existing bug not related to this PR
- [x] Un-snowflake atmobeta when testing complete

- [x] Documentation created/updated (just changelog?)

### DIFF
--- a/service/instance.py
+++ b/service/instance.py
@@ -169,10 +169,8 @@ def stop_instance(esh_driver, esh_instance, provider_uuid, identity_uuid, user,
     """
     _permission_to_act(identity_uuid, "Stop")
     if reclaim_ip:
-        remove_ips(esh_driver, esh_instance, identity_uuid)
+        remove_floating_ip(esh_driver, esh_instance, identity_uuid)
     stopped = esh_driver.stop_instance(esh_instance)
-    if reclaim_ip:
-        remove_empty_network(esh_driver, identity_uuid, {"skip_network":True})
     update_status(
         esh_driver,
         esh_instance.id,
@@ -234,10 +232,8 @@ def suspend_instance(esh_driver, esh_instance,
     """
     _permission_to_act(identity_uuid, "Suspend")
     if reclaim_ip:
-        remove_ips(esh_driver, esh_instance, identity_uuid)
+        remove_floating_ip(esh_driver, esh_instance, identity_uuid)
     suspended = esh_driver.suspend_instance(esh_instance)
-    if reclaim_ip:
-        remove_empty_network(esh_driver, identity_uuid, {"skip_network":True})
     update_status(
         esh_driver,
         esh_instance.id,
@@ -251,9 +247,9 @@ def suspend_instance(esh_driver, esh_instance,
 
 
 # Networking specific
-def remove_ips(esh_driver, esh_instance, core_identity_uuid, update_meta=True):
+def remove_floating_ip(esh_driver, esh_instance, core_identity_uuid, update_meta=True):
     """
-    Returns: (floating_removed, fixed_removed)
+    Returns: nothing
     """
     from service.tasks.driver import update_metadata
     core_identity = CoreIdentity.objects.get(uuid=core_identity_uuid)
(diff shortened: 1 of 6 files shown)


service/tasks/driver.py
```diff
@@ -299,39 +282,12 @@ def clear_empty_ips_for(username, core_provider_id, core_identity_uuid):
...
-                tenant_name)
-            return (num_ips_removed, True)
-        return (num_ips_removed, False)
-    else:
-        celery_logger.info("No Network found. Skipping %s" % tenant_name)
-        return (num_ips_removed, False)
+    return num_ips_removed
 
 
 @task(name="clear_empty_ips")
```
cdosborn: Want to just nuke this

service/instance.py
```diff
@@ -234,10 +232,8 @@ def suspend_instance(esh_driver, esh_instance,
     """
     _permission_to_act(identity_uuid, "Suspend")
     if reclaim_ip:
-        remove_ips(esh_driver, esh_instance, identity_uuid)
+        remove_floating_ip(esh_driver, esh_instance, identity_uuid)
```
cdosborn: With this change, we no longer need the periodic floating ip task to run as often, we could have it run once a week or less, for instances that are nuked outside of atmosphere

service/tasks/driver.py
c-mart: yep

service/instance.py
c-mart: agreed.

c-mart:
Basic testing on atmobeta completed, and changes made following feedback received so far.

I could use help with the dependent PR for Troposphere, to not display private IP address for an inactive instance.

cdosborn:
@c-mart troposphere is using the ip_address field returned from the /api/v2/instances endpoint, which in turn is just returning the ip_address field of the Instance model. I think we should stop setting the ip_address field to the private ip address. The user should never see this anyway.

cdosborn:
Did a little more digging, would try changing ` _find_esh_ip` in core/models/instance.py

PR #611, opened 2018-04-18
Fix volume quota update

cdosborn:
### Problem
Updating a user's volume quota, would return a successful response including the updated quota values. However, the values wouldn't actually be updated in openstack.

### Solution
Pass a tenant id rather than a tenant name in /os-quota-sets/<tenant id>

The current OS api[1] silently fails, reporting a successful update, even when the update did not occur. Tested against all providers in use. Verified that storage is updated from OS pov.

[1] https://developer.openstack.org/api-ref/block-storage/v3/index.html#quota-sets-extension-os-quota-sets


cdosborn:
Targeting this against v32, because this needs to go in for j7m

PR #618, opened 2018-05-16
Fixes for application_to_provider Glance client v1 native (non-iRODS) upload method

c-mart:
## Description

Glance Client v1 uses a different method for uploading image data; this PR accommodates that.

This codepath was previously untested with Glance Client v1 because we were using the "direct iRODS transfer" mode of application_to_provider. We recently stopped using that (because Marana Cloud Glance is now backed by Ceph), and this bug surfaced.

Also, apparently in Glance API v1, it is valid for an image's checksum to be missing even though that image is in active state; we can catch this and log a warning.

### DIFF
--- a/scripts/application_to_provider.py
+++ b/scripts/application_to_provider.py
@@ -512,10 +512,13 @@ def migrate_or_verify_image_data(img_uuid, src_glance_client, dst_glance_client,
                 logging.warn("Warning: image data already present on destination provider but size does not match; "
                              "this may be OK if image was previously migrated with --clean")
         else:
-            if src_img.checksum != dst_img.checksum:
-                logging.warn("Warning: image data already present on destination provider but checksum does not "
-                             "match; this may be OK if image was previously migrated with --clean, or if iRODS "
-                             "transfer was previously used to migrate image")
+            try:
+                if src_img.checksum != dst_img.checksum:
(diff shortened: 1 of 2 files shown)


scripts/application_to_provider.py
```diff
@@ -577,7 +580,11 @@ def migrate_image_data_glance(src_glance_client, dst_glance_client, img_uuid, lo
         logging.debug("Attempting to upload image data to destination provider")
         with open(local_path, 'rb') as img_file:
             try:
-                dst_glance_client.images.upload(img_uuid, img_file)
+                # "Upload" method is different for Glance client v1, than for v2
+                if isinstance(dst_glance_client, glanceclient.v1.client.Client):
```
cdosborn: Sort of minor, did a bit of reading to confirm this, you have the option to do `type(dst_glance_client) is glanceclient.v1.client.Client`. If the v2 client was a subclass of the v1 client, then your check wouldn't work as intended. `isinstance` is typically preferred, because of duck typing (when you want to check if an instance has at minimum the behavior of some class and its parents), but in cases like this I think the preferred approach is to check for an exact class match.

c-mart: sound reasonable, change made as suggested

PR #626, opened 2018-06-18
Django tests fail when run consecutively

cdosborn:
## Description
Fix test failures when run consecutively

## Problem
Running tests twice causes failures

## Solution
Remove the memoize cache before test runs

This would fail if run twice:
```
./manage.py test --keepdb jetstream.test_tas_api.TestJetstream.test_validate_account
```

It would fail because the cassettes did not reflect the correct number of plays, i.e. api calls didn't happen like we expected. Normally cassettes store the results of a first http call, and intercept and respond to all future calls with the results of the first. We assert that the cassette was called a certain number of times. However, django-memoize intercepts the functions that would even make these http calls, so the cassettes were under-reporting the number of playbacks.


cdosborn:
Related https://github.com/cyverse/atmosphere/pull/511

PR #629, opened 2018-06-25
Fix mount failure despite attach success

cdosborn:
## Description
Fix mount failure despite attach success

### Problem
After what appears like a successful volume attach, the volume would not be mounted in the instance

### Solution
Wait to mount until the volume reports "in-use"

</br>
When attaching a volume against Openstack, atmosphere would check if the volume was in attaching. If it was not in that state, it would exit without failure. The chain would continue and the ansible would attempt to mount, despite the volume not being ready. Normally the volume moves immediately to "attaching" however occasionally it would move first to "reserved".

The code has been altered to wait until the volume is "in-use", it will either timeout trying or get there eventually. Additionally, if the volume doesn't report the proper attachments it will raise an exception and not silently fail as before.

It is possible that if the volume is attached elsewhere that this approach will not work. Neither this fix not the original code was designed to handle multiple volume attachments.

This is a snippet of the logs produced from the new code addressing the original error:
```
attach_task started at 2018-06-25 19:01:54.722485+00:00.
Volume ddd7bdb3-2ad3-4998-aa8c-5baa681207f8 is not ready. Expected 'in-use', got 'reserved'
attach_task finished at 2018-06-25 19:02:09.104036+00:00.
Task attach_task[a67dc3a0-0f63-4513-9ebf-e98c53919583] succeeded in 14.4251165979s: '/dev/sdb'
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
   - Explicitly specify external network to rtwo when associating a floating IP
     address ([#624](https://github.com/cyverse/atmosphere/pull/624))
     ([#632](https://github.com/cyverse/atmosphere/pull/632))
+  - Attaching task succeeded before volume was actually attached causing
+    volume mount to fail
+    ([#629](https://github.com/cyverse/atmosphere/pull/629))
 
 ### Removed
   - In the general feedback email we no longer include the users selected


service/tasks/volume.py
```diff
@@ -154,62 +154,49 @@ def mount_volume_task(driverCls, provider, identity, instance_id, volume_id,
 # Libcloud Instance Action (Attachment) tasks
 
 
-@task(name="attach_task",
-      default_retry_delay=20,
-      ignore_result=False,
-      max_retries=1)
```
cdosborn: ignore_result's default is `False`, we don't need to retry so retry delay and max_retries have been removed

service/tasks/volume.py
```diff
@@ -154,62 +154,49 @@ def mount_volume_task(driverCls, provider, identity, instance_id, volume_id,
...
-                            sleep_time))
-            time.sleep(sleep_time)
+        volume_status = volume.extra.get('status', '')
+        if volume_status == "in-use":
+            break
 
-        if 'available' in volume.extra.get('status', ''):
-            raise Exception("Volume %s failed to attach to instance %s"
-                            % (volume.id, instance_id))
+        if attempts > 4:
```
cdosborn: At most we wait 40sec

service/tasks/volume.py
```diff
@@ -154,62 +154,49 @@ def mount_volume_task(driverCls, provider, identity, instance_id, volume_id,
...
+
+    try:
+        attach_data = volume.extra['attachments'][0]
+        device = attach_data['device']
+    except (IndexError, KeyError):
+        raise Exception("Could not find 'device' in volume.extra {}".format(
+            volume.extra))
+
+    celery_logger.debug("attach_task finished at %s." % timezone.now())
+    return device
```
cdosborn: The guarantee with this task is that you either get a device or a raised exception

julianpistorius: 🎉

c-mart: Importantly, that you get a _device on an in-use volume_ or an exception

c-mart:
[OpenStack volume statuses](https://developer.openstack.org/api-ref/block-storage/v3/#volumes-volumes) for reference

PR #262, opened 2017-02-07
Use date extension file naming for logrotate for easy archiving.

julianpistorius:
### Problem

Current logrotate file naming convention is not conducive to archiving.

### Solution

Add a date to the end of the rotated files. (`dateext` option)

refs [#ATMO-1682](https://pods.iplantcollaborative.org/jira/browse/ATMO-1682)

PR #631, opened 2018-06-29
ATMO-2149: Atmosphere deploy without Subspace

calvinmclean:
## Description

Replace Subspace by using Ansible's PlaybookCLI to deploy instances.

The main purpose of this PR was to reduce the amount of core Ansible code that had to be modified by Subspace in order to make it easier to update the program to work with newer Ansible versions. In the end, Subspace was completely removed.

In the place of Subspace, we use Ansible's PlaybookCLI object. The `execute_playbooks()` function adds two attributes to `ansible.constants` (`ansible.cfg` values) to suppress Ansible playbook stdout and set the directory used for logging. This is done here because `ansible.cfg` expects a file path, not a directory path, and would cause errors on regular runs of `ansible-playbook` and we do not want to suppress stdout on regular `ansible-playbook` runs. Then, it loops through playbook files and for each one:
  1. Creates a list of arguments identical to those that would be used with the `ansible-playbook` command (`--inventory-file`, `--limit`, `--extra-vars`, and the playbook path)
  2. Creates the PlaybookCLI object and parses the arguments
  3. Runs the playbook and exits the loop if there is a failure

It returns a list of return codes from each playbook.

More changes were made to the functions that handle the return values to expect integers instead of playbook stats. This greatly simplifies the process and allows for easy upgrades to newer Ansible versions.

Related PRs:
  - [Atmosphere-Ansible PR #154](https://github.com/cyverse/atmosphere-ansible/pull/154)
  - [Clank PR #262](https://github.com/cyverse/clank/pull/262)

### DIFF
--- a/service/tasks/volume.py
+++ b/service/tasks/volume.py
@@ -19,11 +19,10 @@ from atmosphere.settings.local import ATMOSPHERE_PRIVATE_KEYFILE
 from service.driver import get_driver
 from service.deploy import (
     deploy_check_volume, deploy_mount_volume, deploy_unmount_volume,
-    build_host_name, execution_has_failures, execution_has_unreachable)
+    execution_has_failures, execution_has_unreachable)
 from service.exceptions import DeviceBusyException
 
 # Ansible deployment tasks
-
 
 @task(name="check_volume_task",
       max_retries=0,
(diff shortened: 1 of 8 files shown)


service/tasks/volume.py
```diff
@@ -48,14 +48,12 @@ def check_volume_task(driverCls, provider, identity,
         playbooks = deploy_check_volume(
```
cdosborn: I believe this should be playbook_results or something else

service/tasks/volume.py
```diff
@@ -48,14 +48,12 @@ def check_volume_task(driverCls, provider, identity,
...
             instance.ip, username, instance.id,
             device_location, device_type=device_type)
-        celery_logger.info(playbooks.__dict__)
         hostname = build_host_name(instance.id, instance.ip)
-        result = False if execution_has_failures(playbooks, hostname)\
-            or execution_has_unreachable(playbooks, hostname) else True
+        result = False if execution_has_failures(playbooks) or execution_has_unreachable(playbooks) else True
```
cdosborn: playbooks is a misnomer

service/tasks/volume.py
```diff
@@ -48,14 +48,12 @@ def check_volume_task(driverCls, provider, identity,
...
         if not result:
             raise Exception(
                 "Error encountered while checking volume for filesystem: %s"
-                % playbooks.stats.summarize(host=hostname))
+                % playbooks)
```
cdosborn: Probably don't want to log this, should instead log instance_id and volume_id

calvinmclean:
Whoops I missed a few of those. I went through and found the rest and removed some function calls that were unnecessary (like building a hostname that is never used)

cdosborn:
I tested this and launched two instances and confirmed that deployment/volume attach/volume detach work.

Again great work.

PR #641, opened 2018-08-02
Fix issue where rtwo pagination resulted in failing to retrieve all instances

cdosborn:
See https://github.com/cyverse/rtwo/pull/31


cdosborn:
I manually tested this against all providers and assured that the correct number of instances were reported. This change accounts for the fact that IU is not returning consistent API responses. If we run into further issues, we'll press for that issue to be resolved.

PR #644, opened 2018-08-06
Only return identities that exist on active providers

cdosborn:
### Problem
The volume creation modal was allowing users to choose providers/(really identities) that were no longer active

### Solution
Only return idents for active and non-endated providers.

- [ ] Reviewed and approved by at least one other contributor


cdosborn:
I did a little bit of exploring here. It's possible this isn't the right fix and the frontend should be doing some client side filtering, but I'm indifferent. Troposphere doesn't handle inactive providers in at least another obvious way. When you go to launch an instance you can select applications that only exist on *inactive* providers.

PR #646, opened 2018-08-10
Configure atmosphere to optionally report allocation usage

cdosborn:
## Description
This was created so that a staging environment could include production TAS credentials without worrying about having an impact on reporting. The staging and production environment would run concurrently at times and we wouldn't want to overreport usage.

In the past we used the development tas api. This is just another route that allows us to change less variables between production/staging and avoid requests to grant users allocations on both production tas and developer tas.

https://github.com/cyverse/clank/pull/266

### DIFF
--- a/jetstream/tasks.py
+++ b/jetstream/tasks.py
@@ -138,6 +138,11 @@ def _create_tas_report_for(user, tacc_username, tacc_project_name, end_date):
 
 @task(name="report_allocations_to_tas")
 def report_allocations_to_tas():
+    if not settings.REPORT_ALLOCATION_USAGE:
+        logger.warn(
+            "Report sending is disabled! The TACC API will NOT see changes in allocation ({})".
+            format(settings.TACC_API_URL))
+        return
     logger.info("Reporting: Begin creating reports")
     create_reports()
     logger.info("Reporting: Completed, begin sending reports")
--- a/atmosphere/settings/local.py.j2
+++ b/atmosphere/settings/local.py.j2
@@ -38,6 +38,12 @@ ENFORCING = True
 ENFORCING = False
 {% endif %}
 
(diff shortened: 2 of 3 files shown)


jetstream/allocation.py
```diff
@@ -122,6 +122,11 @@ def report_project_allocation(self, report_id, username, project_name, su_total,
         if not type(su_total) in [int, float]:
             raise Exception("SU total should be integer or float")
 
+        # Abort here to prevent reporting of usage, should abort in all
+        # development environments to prevent double reporting
+        if not settings.REPORT_ALLOCATION_USAGE:
+            return True
```
julianpistorius: I think this should probably return False, not True. If this returns True the calling code will mark the report as sent.

julianpistorius: And/or use this flag higher up in the call tree.

cdosborn: I wanted it to mark the report as sent, so that the app wouldn't continue to accumulate reports. Essentially this flag will convince atmo that the reports were received. Its sort of akin to mocking the tas api to return a 200 on sending them. However I could see it leaving someone puzzled that the db flags them as sent, but the reports were never actually sent.

I'll move it up so the flag makes the `report_allocations_to_tas` a no-op.

jetstream/tasks.py
```diff
@@ -138,6 +138,11 @@ def _create_tas_report_for(user, tacc_username, tacc_project_name, end_date):
 
 @task(name="report_allocations_to_tas")
 def report_allocations_to_tas():
+    if not settings.REPORT_ALLOCATION_USAGE:
+        logger.warn(
+            "Report sending is disabled! The TACC API will NOT see changes in allocation ({})".
+            format(settings.TACC_API_URL))
+        return
```
julianpistorius: 👍

PR #648, opened 2018-08-13
Add new API endpoint to allow for create/edit/delete of tokens by user

calvinmclean:
## Description

This PR adds an API endpoint that allows users to create/edit/delete named tokens that they can use to access Atmosphere from something like the Atmosphere CLI.
The tokens are a new model that extends the existing cyverse-django-auth tokens and adds a name.
The tokens have a 5 year expiration date.

~~I still need to create the migration for this!~~

[Related Troposphere PR](https://github.com/cyverse/troposphere/pull/789)

- [x] If creating/modifying DB models which will contain secrets or sensitive information, PR to [clank](https://github.com/cyverse/clank) updating sanitation queries in `roles/sanitary-sql-access/templates/sanitize-dump.sh.j2`
    - It was already [in there](https://github.com/cyverse/clank/blob/master/roles/sanitary-sql-access/templates/sanitize-dump.sh.j2#L39) because the `access_token` table existed but was empty/unused


cdosborn:
It looks like django cyverse auth doesn't look at the expirationTime of the token when it does token auth

cdosborn:
Api view needs to return expired tokens. I believe these are filtered out.

calvinmclean:
I got the tests working just a few minutes before I have to leave today so I will polish and commit them when I am in the office on Wednesday!

cdosborn:
Was talking with Tharon, no need to have these tokens expire.

calvinmclean:
Removed token expireTime and finished the tests :+1:

cdosborn:
Creation of the d-c-a Token should pass an "issuer"

core/models/access_token.py
```diff
@@ -16,7 +16,7 @@ class Meta:
         app_label = "core"
 
 def create_access_token(user, token_name=None, token_expire=None, remote_ip=None, issuer=None):
-    token = Token(user=user)
+    token = Token(user=user, issue=issuer)
```
cdosborn: issue -> issuer

calvinmclean: WHOOPS!

api/v2/views/access_token.py
```diff
@@ -0,0 +1,45 @@
...
+        """
+        user = self.request.user
+        qs = AccessToken.objects.filter(token__user=user)
+        if 'archived' in self.request.query_params:
+            return qs
+        # Return current results
+        return qs.filter(only_current_access_tokens())
+
+    def create(self, request):
+        issuer_backend = request.session.get('_auth_user_backend', '').split('.')[-1]
```
cdosborn: Curious, why are you looking this up in the session?

cdosborn: Issuer may just as well be "Personal-Access-Token"

calvinmclean: It is the only way I knew how to get the issuer by copying what is done in [`auth.py`](https://github.com/cyverse/atmosphere/blob/468a1cc3254efb0a793a4254112d6e3acf20c518/api/auth.py#L50-L53)

cdosborn:
A user shouldn't be allowed to create access tokens with the same name

api/v2/views/access_token.py
cdosborn: In this case you're the issuer. Do some searching for where we pass values for issuer in d-c-a, which should show its purpose.

cdosborn: The session is a key value store which is used in web-apps to track a user across multiple stateless requests. It's a bag of properties. If a user submits requests and passes their session as a cookie, we'll know which session they're using. `_auth_user_backend` is probably some relic of long ago

calvinmclean: Ok, I will look into that

calvinmclean:
I was able to whip up some changes to disallow creating tokens with the same name. I just need to improve the error message, make some tests, and make the query only look for tokens with the same name AND same user (I didn't implement the user part yet). The change you mentioned in Tharon's PR about having the NotificationController show the error from the API will show the error message I am creating here. 

Unfortunately I can't finish this up until Friday. Class is so inconvenient 😢

cdosborn:
I was thinking about this some more. There's nothing technically wrong with a user creating the same token from the api's perspective. Each token has a unique id, and the api response includes said ids. We don't really care if users do this for projects etc.

calvinmclean:
In response to "Api view needs to return expired tokens. I believe these are filtered out.":

Should I still do this now that the tokens do not expire?

calvinmclean:
I created the commit for enforcing unique names because I had it mostly finished anyways so now we can look at that when deciding what to do. However, users are still able to rename tokens to have identical names because, to be honest, I am not sure how the editing works on the API side of this. Maybe we need to enforce that on the Troposphere side (in collections or stores)?

cdosborn:
When you're enforcing some guarantee you want to enforce at the lowest level possible, a db constraint. If you can't enforce there, then you enforce at the api level, and hope your logic covers the edge cases (ex. you may have forgotten to replace update logic you mentioned above). The last place to add a constraint is in the ui, which is the weakest place, because no one has to use the frontend and they could just as easily change it to have different behavior. 

A db constraint can only exist on a given table. You would have to add a user field to your token, and then you could add unique constraint between (user, name). This would imply that those two columns MUST uniquely identify a record, the db wouldn't allow any inserts/updates/changes which would break that guarantee. This isn't great because the token model has a user. So your access_token.user and access_token.token.user would be redundant and would open up potential for bugs. One path here would be to remove the user field from token, a quick glance showed that it was unnecessary.

Or you could just add it in the api, but none of our other apis do this checking. In each of those scenarios if the user is confused by making a duplicate, they can feel pretty responsible for doing something like that and not blame the application (so the weight isn't that important).

A middle ground solution would be to allow duplicate names, but to show the creation date of the token. With this information users could likely figure out which duplicate was which.

cdosborn:
I don't think we need to worry about this issue. I still have feedback on your pr.

cdosborn:
Your serializer isn't being used in your create method which is a mistake.

cdosborn:
agh accidentally closed

cdosborn:
Re: https://github.com/cyverse/atmosphere/pull/648#issuecomment-415837627
Yes. If the tokens do not expire, it is non-sensical to filter them. If they do expire at some point, we will also want the filter to be removed.

cdosborn:
Another reason to enforce constraints at the database level. The python api code runs in parallel. So the code you added says something like "If the token exists already with this name, then error". However if two requests to create a token with the same name happened in parallel, they would both be created with the same name.

The database constraint would prevent this.

core/models/access_token.py
```diff
@@ -0,0 +1,24 @@
+from django.db import models
+from django.utils import timezone
+
+from django_cyverse_auth.models import Token
+
+
+class AccessToken(models.Model):
+    """
+    Extend the django_cyverse_auth Token to add a name
```
cdosborn: This is literally describing what the model is, may explain how this is used in the context of the api.

core/models/access_token.py
```diff
@@ -0,0 +1,24 @@
...
+    name = models.CharField(max_length=128, null=False, blank=False)
+
+    class Meta:
+        db_table = "access_token"
+        app_label = "core"
+
+def create_access_token(user, token_name=None, token_expire=None, remote_ip=None, issuer=None):
+    if AccessToken.objects.filter(name=token_name, token__user_id=user.id):
+        return None
+    token = Token(user=user, issuer=issuer)
```
cdosborn: This line and one below can be replaced by a single ```token = Token.objects.create(...)```

core/models/access_token.py
```diff
@@ -0,0 +1,24 @@
...
+    """
+    token = models.OneToOneField(Token, on_delete=models.CASCADE)
+    name = models.CharField(max_length=128, null=False, blank=False)
+
+    class Meta:
+        db_table = "access_token"
+        app_label = "core"
+
+def create_access_token(user, token_name=None, token_expire=None, remote_ip=None, issuer=None):
+    if AccessToken.objects.filter(name=token_name, token__user_id=user.id):
```
cdosborn: ```AccessToken.objects.filter(name=token_name, token__user=user)```

core/models/access_token.py
```diff
@@ -0,0 +1,24 @@
...
+        return None
+    token = Token(user=user, issuer=issuer)
+    token.save()
+    access_token, created = AccessToken.objects.update_or_create(token=token, name=token_name)
```
cdosborn: This may have been a recent change, but you don't want `update_or_create`. You just want create, a new token should always be returned, since a new token is always created.

core/models/access_token.py
calvinmclean: Would this still check if a token with that name already exists? Or does it rely on the database enforcing that constraint?

cdosborn: I'm referring to line 21, 22 here

calvinmclean: Oh okay. To make the name enforced by the database, I can probably add `unique=True` to the CharField and then I don't need 19 and 20 either

cdosborn: That would mean make the name unique across all tokens, so two users couldn't have token 'foobar'

core/migrations/replace_0001_initial.py
```diff
@@ -0,0 +1,1466 @@
...
+import django.utils.timezone
+import uuid
+
+
+# Functions from the following migrations need manual copying.
+# Move them and any dependencies into this file, then update the
+# RunPython operations to refer to the local versions:
+# core.migrations.0088_set_project_leaders_and_authors
+# core.migrations.0089_projects_to_project__alter_created_by
+# core.migrations.0092_set_unknown_size_instance_status_history
```
cdosborn: Would you replace this comment with a short summary of why this migration exists.

calvinmclean:
I created a `validate` method in the serializer following [django docs](https://www.django-rest-framework.org/api-guide/serializers/#object-level-validation)

calvinmclean:
Since the validate method returns the valid data, maybe I should use `data = AccessTokenSerializer().validate(request.data)` in the `create` method, but it does not seem necessary since the data is not changed at all

api/v2/views/access_token.py
```diff
@@ -0,0 +1,37 @@
...
+    serializer_class = AccessTokenSerializer
+
+    def get_queryset(self):
+        """
+        Filter projects by current user.
+        """
+        return AccessToken.objects.filter(token__user=self.request.user)
+
+    def create(self, request):
+        AccessTokenSerializer().validate(request.data)
```
cdosborn: Try replacing this line with:
```
++        serializer = self.get_serializer(data=request.data)
++        serializer.is_valid(raise_exception=True)
```

Your class is implementing an AuthModelViewSet, which is based off of rest_framework's ModelViewSet, which provides a default implementation for `def create`. If you're going to override it, you just need to know what all its supposed to do. It basically takes `serializer_class` above does some shit and returns a json repsonse from the serializers data. You don't have to implement the serializers validate method. That method is based on the serializers fields (which in turn are based on the fields defined on your model).

In a manage.py shell run:
```
from api.v2.serializers.details import AccessTokenSerializer
AccessTokenSerializer()
```

cdosborn: A helpful trick when you're replacing a method:
```
def create(self, request):
    import ipdb; ipdb.set_trace()
    super(self.__class__, self).create(request)
```
That will call the superclass implementation and you can step through and see what it does.

api/v2/views/access_token.py
```diff
@@ -22,7 +22,8 @@ def get_queryset(self):
         return AccessToken.objects.filter(token__user=self.request.user)
 
     def create(self, request):
-        AccessTokenSerializer().validate(request.data)
+        serializer = self.get_serializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
```
cdosborn: :+1: May seem a tad overkill for so simple an api, but this future proofs as more fields get added

calvinmclean: Better to do it right the first time for sure. I appreciate all the feedback, I am learning a lot about APIs!

calvinmclean:
Once this is approved, I would like a chance to rebase and squash some of these commits before merge

cdosborn:
Troposphere expects error responses to be valid json. 

For example this is what rest_framework returns when i send gobbledygook:
```
$ curl -k -X POST 'https://local.atmo.cloud/api/v2/access_tokens' -H "Content-Type: application/json" -H 'Accept: application/json' -H 'Authorization: Token ...' --data '{ {} }'      
{"detail":"JSON parse error - Expecting property name enclosed in double quotes: line 1 column 3 (char 2)"}
```

A raw string is not technically a json document, and tropo expects to see one.

calvinmclean:
Oh yeah, there is still one problem: token names can be changed to match existing names without error. Any suggestions to fix this?

api/tests/v2/test_access_tokens.py
```diff
@@ -80,7 +80,7 @@ def test_create_same_name(self):
         force_authenticate(self.create_request, user=self.user)
         response = self.create_view(self.create_request)
         self.assertEquals(response.status_code, 400)
-        self.assertEquals(response.data, "Token with name \"Test Token Creation\" exists.")
+        self.assertEquals(response.data, {'detail': u'Token with name "Test Token Creation" exists.'})
```
cdosborn: :+1:

cdosborn:
It would be a little more work, but we could give up on checking if the names are unique, and instead just include a creation date (the serializer would pull this field off the Token)

cdosborn:
The ui could list each token with its creation date, and the issue would no longer exist

cdosborn:
I sort of like that route, because in the absence of an expiration date, it would be nice to show how old a token is

cdosborn:
Another route would be to perform the validation in the update method

cdosborn:
I'll wip this up real quick

cdosborn:
@calvinmclean feel free to rebase, squash, make sure you refetch this branch, its history has been altered

calvinmclean:
Rebase complete for beautiful git history 👍 
However, I tried to drop the 'Update CHANGELOG' to be the last commit, but it doesn't show up like that here. It does show as the last commit with `git log` though

PR #656, opened 2018-08-31
Remove code related to old allocation system

cdosborn:
## Description

Remove code related to old allocation system. There was a lot of code which depended on code which depended on the old allocation. The leaves of this dependency tree (i.e. didn't depend on anything) were removed first. Each commit unfurls the next layer so to speak, so that any point in time I was just removing code not being used.


cdosborn:
how does the coverage drop by a fraction of a percent when you remove 3K lines :confused:

PR #658, opened 2018-09-05
Prefer importing settings from django.conf

cdosborn:
## Description

The proper way to import settings is via 'from django.conf import settings'. This is because settings are a part of Django's abstractions. For example, `@override_settings` has no effect unless settings are imported this way.

https://docs.djangoproject.com/en/1.10/topics/settings/#using-settings-in-python-code

### DIFF
--- a/core/email.py
+++ b/core/email.py
@@ -14,7 +14,7 @@ from pytz import timezone as pytz_timezone
 
 from threepio import logger
 
-from atmosphere import settings
+from django.conf import settings
 from core.models.allocation_source import total_usage
 from core.models import IdentityMembership, MachineRequest, EmailTemplate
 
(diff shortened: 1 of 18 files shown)


core/email.py
```diff
@@ -14,7 +14,7 @@
 
 from threepio import logger
 
-from atmosphere import settings
```
calvinmclean: According to the Travis tests, line 154 of this file is causing the error because of `settings._get_method_for_string` which is a function defined [here](https://github.com/cyverse/atmosphere/blob/ce15e0f7b77394e37470af5ff959ca2e7679e73e/atmosphere/settings/__init__.py#L540). I think it can be replaced by just changing the line to `lookup_fn = globals()[lookup_fn_str]`

calvinmclean: I made this change and ran the tests in my dev environment and the tests passed.

The method is used again on [line 166 of the same file](https://github.com/cyverse/atmosphere/blob/ce15e0f7b77394e37470af5ff959ca2e7679e73e/core/email.py#L166)

PR #660, opened 2018-09-06
Fix incorrect order of arguments

cdosborn:
## Description

Incorrect order of arguments would result in an exception always being raised. So the original intention of allowing validation to work in the absence of the TAS api was never realized.


cdosborn:
Would like to write a test to validate that this behavior is actually fixed

PR #662, opened 2018-09-07
Introduce vulture to eliminate dead code

cdosborn:
## Description
This pr used vulture to identify some dead code. Travis has been updated to check builds for dead code. Vulture is tuned to raise warnings when it is 60% confident that code is dead.

There is still a lot that is dead but below the 60% threshold, so this is just a starting place. I added some documentation for how we're using vulture [here](https://github.com/cyverse/atmosphere/blob/5cf02207cc2c6b213d240c0f04872201de607f6a/travis/check_for_dead_code_with_vulture.sh#L3-L18).


c-mart:
Please don't remove scripts/add_new_provider.py:

https://github.com/cyverse/atmosphere-guides/blob/master/src/connecting_cloud_provider/connecting_cloud_provider.md#add-a-new-provider-

cdosborn:
Thanks @c-mart

PR #664, opened 2018-09-07
Make linter significantly more strict

cdosborn:
## Description

Increase linter from "low" to "medium"

Each commit fixes a single lint violation. I don't expect a reviewer to look at each commit. There are only a handful of commits that change more than one line, and they all fell into the same category of issue relating to imports. 

File A, would `from foo import X`, but A would never use X at all. So the linter would suggest to remove that import. However file B, would import `X from A`. So file B would have to be updated to also import from `foo`. A file should import a value from the package where the value is defined. 

### DIFF
--- a/service/instance.py
+++ b/service/instance.py
@@ -1532,7 +1532,6 @@ def user_network_init(core_identity):
     username = core_identity.get_credential('key')
     if not username:
         username = core_identity.created_by.username
-    esh_driver = get_cached_driver(identity=core_identity)
     dns_nameservers = core_identity.provider.get_config('network', 'dns_nameservers', [])
     subnet_pool_id = core_identity.provider.get_config('network', 'subnet_pool_id', raise_exc=False)
     topology_name = core_identity.provider.get_config('network', 'topology', raise_exc=False)
--- a/service/instance.py
+++ b/service/instance.py
@@ -1697,7 +1697,6 @@ def _update_instance_metadata(esh_driver, esh_instance, data={}, replace=False):
     NOTE: This will NOT WORK for TAGS until openstack
     allows JSONArrays as values for metadata!
     """
-    wait_time = 1
     if not esh_instance:
         return {}
     instance_id = esh_instance.id
--- a/service/instance.py
+++ b/service/instance.py
@@ -1656,7 +1656,6 @@ def _extra_openstack_args(core_identity, ex_metadata={}):
         # FIXME: In a new PR, allow user to select the keypair for launching
         user_key = user_keys[0]
         ex_keyname = user_key.name
-    security_group_name = core_identity.provider.get_config("network", "security_group_name", "default")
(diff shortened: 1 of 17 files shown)


cdosborn:
lol, at the top you'll see there are 464. I don't expect you to review each one. Look at the handful of commits that touch more than one file (or more than one line), maybe try and understand those. Or look at the commit messages and find any linting errors you don't understand and look at those.

service/instance.py
```diff
@@ -743,11 +723,8 @@ def update_status(esh_driver, instance_id, provider_uuid, identity_uuid, user):
     if esh_driver.provider.location.lower() == 'mock':
         return None
     # Convert & Update based on new status change
-    core_instance = convert_esh_instance(esh_driver,
-                                         esh_instance,
-                                         provider_uuid,
-                                         identity_uuid,
-                                         user)
+    convert_esh_instance(esh_driver, esh_instance, provider_uuid, identity_uuid,
```
calvinmclean: The linter made this change because this used to be assigned to variable `core_instance`, but that variable is never used. That makes sense, but I am just commenting because it also changed the multi-line syntax of the function call. Which way is preferred? It seems weird to me to just put one of the parameters on its own line, but I also dislike having each on its own line. Is something like this bad form?
```
convert_esh_instance(esh_driver, esh_instance, provider_uuid,
                     identity_uuid, user)
```

This is a little crazy but I kinda like it:
```
args = (esh_driver, esh_instance, provider_uuid, identity_uuid, user)
convert_esh_instance(*args)
```

cdosborn: From my perspective, the style doesn't matter as much as consistency. There are some styles that feel better than others, but then occasionally you'll read some code and you'll just wish that it would have chosen **any** style, because its illegible.

So the `*args` makes for good use of space, but I'd prefer to leave style to a program that is pretty darn good at ensuring legibility and also eliminates any time spent debating personal aesthetic. *args and **kwargs have a place but they make for easy mistakes, like switching the param order, or not having enough args. Or the person who changes all instances of `convert_esh_instance` to the new interface doesn't see the code that built up a *args list.

My next work will be to format the codebase.

calvinmclean: Thank you for clarifying, I completely agree. I recently learned about using `*args` with a tuple so I just like doing it 🤓

api/v1/views/meta.py
```diff
@@ -40,18 +40,9 @@ def get(self, request, provider_uuid, identity_uuid):
         if not esh_driver:
             return invalid_creds(provider_uuid, identity_uuid)
         data = add_user_urls(request, provider_uuid, identity_uuid)
-        if request.user.is_staff:
-            add_staff_urls(request, provider_uuid, identity_uuid)
```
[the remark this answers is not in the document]
cdosborn: If you peak inside `add_user_urls` you'll see that it's not doing anything. If it had had some important side effect I would have kept the function call. The function was also not used anywhere else

PR #665, opened 2018-09-10
ATMO-2169: Fix CC admins to expect only one admin

calvinmclean:
## Description

~Problem: ADMINS variable used for admins emails is set up to allow multiple admins, but this is not used.~

~In order to fix this, the ADMINS variable is changed to ADMIN and changed to be a single tuple. The uses of it are also changed to expect this new type.
Other variables ATMO_DAEMON, SUPPORT_EMAIL, and ATMO_SUPPORT that are derived from ADMINS are changed also.~

~If we prefer to still email multiple admins, I have changes that would make that happen in [this commit](https://github.com/calvinmclean/atmosphere/commit/f53febb10f8a4bb1ef60f4a56b73a4af4c5e5df2). However, this seems like overkill in most situations since reported instances get brought to all of our attention through Intercom anyways.~

Problem: Email in Atmosphere is difficult to understand and use because of the variables and methods

'email_to_admin' was being used to send emails to support or to admins depending on the `request_tracker` boolean so it was replaced by merging it into 'email_admin' and creating 'email_support'.

Refactored variables:
  MANAGERS removed because it is used by Django for a feature we don't leverage
  ATMO_DAEMON now sends emails and is always 'atmo-notify@SERVER_NAME'
  ATMO_SUPPORT is now a single email address to receive support requests instead of a list
  SUPPORT_EMAIL is removed because it was only used to tell users who to email (replaced by ATMO_SUPPORT)
  ADMINS remained the same but the first entry is no longer used to send emails

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,11 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
   - Format codebase with `yapf` require code to be formatted in travis build
     ([#677](https://github.com/cyverse/atmosphere/pull/677))
 
+### Changed
+  - Refactored email to make variables and methods used for sending emails
+    easier to understand and use
+    ([#665](https://github.com/cyverse/atmosphere/pull/665))
+
 ### Fixed
   - Fix nginx gateway timeout on unshelve
     ([#686](https://github.com/cyverse/atmosphere/pull/686))


core/email.py
```diff
@@ -239,11 +239,9 @@ def email_to_admin(
...
     elif not username:  # user_email provided
         username = 'Unknown'
-    if request_tracker or not cc_user:
-        # Send w/o the CC
-        cc = []
-    else:
-        cc = [email_address_str(username, user_email)]
+    cc = None
+    if not request_tracker and cc_user:
+        cc = email_address_str(username, user_email)
```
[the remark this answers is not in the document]
calvinmclean: I'm not sure what the problem in ATMO-2169 is then since it is already CC'ing a list of one email and the ADMINS_EMAIL_TUPLE variable is not used in Atmosphere.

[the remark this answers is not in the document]
cdosborn: ADMINS_EMAIL_TUPLE is used in Atmosphere. The ticket's purpose (unclear) was to make things more reasonable. Currently on j7m prod we configure that variable with a list of admins emails, but only the first is used. We don't strictly need the multiple admins there, so the ticket was open to the option of either supporting a list of admins or a single admin, but that the variables needed to reflect the actual use in the code.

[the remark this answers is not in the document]
calvinmclean: Oh I see, ADMINS_EMAIL_TUPLE is used to define the ADMINS variable.

calvinmclean:
I just looked at the variables used by Clank, and `ADMINS_EMAIL_TUPLE` and `ATMO_SUPPORT_EMAIL_TUPLE` are actually lists, not tuples, so I will have to change those when I am back in the office next, or change how the j2 file parses those to make them into tuples. Probably better to change them in the Clank variables files because they are called tuples.

atmosphere/settings/__init__.py
```diff
@@ -41,16 +41,14 @@
 ALLOWED_HOSTS = [unicode(SERVER_URL.replace('https://', ''))]
 
 # NOTE: first admin will be sender of atmo emails.
-ADMINS = (
-    ('AT LEAST ONE ADMIN REQUIRED', 'sends-email@if-debug-false.com'),
-)
+ADMIN = ('AT LEAST ONE ADMIN REQUIRED', 'sends-email@if-debug-false.com')
```
cdosborn: Comment on 43 can be nuked, and first value of dummy ADMIN no longer makes sense

atmosphere/settings/__init__.py
```diff
@@ -41,16 +41,14 @@
...
 
 # Set this value to be different if you want to separate where 'support' related email is sent.
-ATMO_SUPPORT = ADMINS
-ATMO_DAEMON = (("Atmosphere Daemon", "atmo-alerts@iplantcollaborative.org"),)
+ATMO_SUPPORT = ADMIN
+ATMO_DAEMON = ("Atmosphere Daemon", "atmo-alerts@iplantcollaborative.org")
 
 # Django uses this one..
-MANAGERS = ADMINS
+MANAGERS = ADMIN
```
cdosborn: Can you look at what `MANAGERS` is used for, this seems unlikely to work if it previously expected an iterable

core/email.py
```diff
@@ -51,30 +51,6 @@ def email_address_str(name, email):
...
-                return admin_user, admin_email
-    return (settings.ADMINS[0][0], settings.ADMINS[0][1])
-
-
-def atmo_daemon_address():
-    """ Return the daemon email address.
-    """
-    return (settings.ATMO_DAEMON[0][0], settings.ATMO_DAEMON[0][1])
-
-
```
cdosborn: Glad to get rid of these methods

atmosphere/settings/__init__.py
calvinmclean: Looks like [`MANAGERS` expects a list](https://docs.djangoproject.com/en/1.9/ref/settings/#std:setting-MANAGERS). Also, this section of the documentation mentions that `ADMINS` is also a variable that django expects to be a list. However, it does not seem like either are being used by django unless `ADMINS` are getting emails for all exceptions.

calvinmclean: Also, Atmosphere defines `ADMINS` as a list of lists when Django expects a list of tuples so this is more evidence that Django hasn't been using it

cdosborn: Yeah both are iterable is the important bit

cdosborn: That doc is pretty specific, if we don't use that middleware, we should remove `MANAGERS`

cdosborn: I just read the doc for `ADMINS` seems like the right way to go. Basically all ADMINS get notified, and the `EMAIL_HOST_USER` sends the emails. Our app will be all the simpler if we can just use Django conventions. 

What do you think about keeping `ADMINS` a list of tuples, and sending admin emails as `settings.EMAIL_HOST_USER`.

calvinmclean: I like that 👍

core/email.py
```diff
@@ -214,19 +190,16 @@ def email_to_admin(
...
-    elif request_tracker:
-        sendto, sendto_email = request_tracker_address()
+    if request_tracker:
+        sendto = settings.ATMO_SUPPORT
     else:
-        sendto, sendto_email = admin_address()
+        sendto = settings.ADMINS
     # E-mail yourself if no users are provided
     if not username and not user_email:
         username, user_email = sendto, sendto_email
```
calvinmclean: I noticed that my changes break this.
How are we sending emails _from_ the user though? Should I just set `from_email=settings.EMAIL_HOST_USER`?

cdosborn: Yeah that would be one way. There may be a way in django. You could probably find how they do it by searching for EMAIL_HOST_USER there

cdosborn: What do you mean from the user? I don't think we do that.

calvinmclean: Okay, I'll dig deeper into that. So far I think we should scrap this whole part:
```
    # E-mail yourself if no users are provided
    if not username and not user_email:
        username, user_email = sendto, sendto_email
    elif not user_email:  # Username provided
        # TODO: Pass only strings, avoid passing 'User' object here.
        if isinstance(username, User):
            username = username.username
        user_email = lookupEmail(username)
        if not user_email:
            user_email = "%s@%s" % (username, settings.DEFAULT_EMAIL_DOMAIN)
    elif not username:  # user_email provided
        username = 'Unknown'
```
and just use `sender=settings.EMAIL_HOST_USER` because we cannot send emails from users' emails, and even if we could, I don't see a benefit. I will see what I think after I look into it more

calvinmclean: This `email_to_admin` function has `from_email=email_address_str(username, user_email)`
I looked some more and most uses of this function use `settings.ATMO_DAEMON` as the user argument, but `send_image_request_failed_email` function inputs atmo username and email

calvinmclean: I think I need to look into this a lot more before I can make any decisions and I don't want to be saying things that I don't know much about. I think I can refactor a lot more than I have so far

calvinmclean:
If you look at a resource request on Intercom and click the email icon under the message, you can see the raw email. The 'From' section shows that the email comes from the email on the user's account

cdosborn:
Good find. It dawned on me that no matter what django is sending the mail with settings.EMAIL_HOST_USER. That user is just used to connect to the smtp server. It's not the from address in mail that is sent. Does that sound right? If the from header is arbitrary, we unlikely want an EMAIL_HOST_USER setting and we just choose something that is non configurable like "atmosphere-admin" the domain of the particular server should distinguish them.

calvinmclean:
Yeah I think it is correct that EMAIL_HOST_USER is used for smtp connection.
I'm not sure I completely understand the rest of the comment. So we shouldn't use EMAIL_HOST_USER at all? 
Are you saying that we should set
```
sender = "Atmosphere Admin <atmosphere-admin@{}>".format(settings.DEFAULT_EMAIL_DOMAIN)
```
(or some other setting that has the host domain)
I kind of like that idea because it reduces the number of variables that we have to define, but it's not very pretty.

cdosborn:
Yeah,  youd want the domain of the host and not the domain of the smtp server. Maybe nuke DEFAULT_EMAIL_DOMAIN and use the hostname. We now correctly set the hostname for the server (i'm guessing there is a django setting for this).

cdosborn:
It looks like the `ATMO_DAEMON` is the one doing the sending. I think we could replace all occurences of ADMINS[0] with sending from ATMO_DAEMON, but make that address like yours above "Atmosphere Admin"

cdosborn:
The latest clank will install an smtp server in your dev env, and correctly set your hostname inside your box. If you havn't already set up an smtp server that could help you test.

atmosphere/settings/local.py.j2
```diff
@@ -238,6 +233,9 @@ DATABASES = {
 SERVER_NAME = SERVER_URL.replace('https://','')
 EMAIL_SUBJECT_PREFIX = "[Django - %s] " % (SERVER_NAME,)
 
+# Sends all emails
+ATMO_DAEMON = ("Atmosphere Admin", "atmo-admin@{}".format(SERVER_NAME))
+
```
cdosborn: This is defined in two places, should probably move into __init__.py. ideally we only have one settings file!

calvinmclean: I think it needs to be in local.py because it uses the SERVER_NAME that is defined here

cdosborn: Okay thats fine

calvinmclean:
The current state of this branch has three email vars:
- `ADMINS`: This variable is unchanged, but it is no longer used to send emails. It is only used in the `email_to_admin` method if `request_tracker` is false.
- `ATMO_DAEMON`: This variable is now used to send all emails
- `ATMO_SUPPORT`: This variable is now a string instead of a list of tuples. It is used as the target in `email_to_admin` when `request_tracker` is true. It is also used in the body of the email sent by `send_instance_email` and as the email referenced in `AllocationBlacklistedError`

Some variables are removed:
- `SERVER_EMAIL`: Never used
- `SUPPORT_EMAIL_SIGNATURE`: Never used
- `MANAGERS`: Never used
- `SUPPORT_EMAIL`: Too similar to `ATMO_SUPPORT` so I removed it. It was only used in the body of the email from `send_instance_email` and in a `AllocationBlacklistedError` exception

cdosborn:
It looks like `ATMO_SUPPORT` and `SUPPORT_EMAIL` we're both target addresses (To: rather than From:).

cdosborn:
`SUPPORT_EMAIL` is like a public facing email address, we tell users to use to reach out to us.

cdosborn:
I think these changes are good. Just confirming with Jetstream that is okay, since they do differentiate between `ATMO_SUPPORT` and `SUPPORT_EMAIL`

atmosphere/settings/__init__.py
```diff
@@ -38,17 +38,12 @@
...
+ADMINS = [('AT LEAST ONE ADMIN REQUIRED', 'receives-email@if-debug-false.com')]
 
 # Set this value to be different if you want to separate where 'support' related email is sent.
-ATMO_SUPPORT = ADMINS
-ATMO_DAEMON = (("Atmosphere Daemon", "atmo-alerts@iplantcollaborative.org"),)
-
-# Django uses this one..
-MANAGERS = ADMINS
+ATMO_SUPPORT = "Atmosphere Support Team <intercom-atmo@cyverse.org>"
+ATMO_DAEMON = "Atmosphere Admin <atmo-notify@cyverse.org>"
```
cdosborn: For consistency, these should be tuples

atmosphere/settings/__init__.py
```diff
@@ -38,17 +38,8 @@
...
-)
-
-# Set this value to be different if you want to separate where 'support' related email is sent.
-ATMO_SUPPORT = ADMINS
-ATMO_DAEMON = (("Atmosphere Daemon", "atmo-alerts@iplantcollaborative.org"),)
-
-# Django uses this one..
-MANAGERS = ADMINS
+# People to notify
+ADMINS = [('AT LEAST ONE ADMIN REQUIRED', 'receives-email@if-debug-false.com')]
```
cdosborn: This can also be removed

cdosborn:
Several ideas for a follow up. 

1. `email_to_admin` and `email_admin` should not both exist. I would change it so there is only `email_admin`.
2. `email_admin` should only email the admin, if you want to send an email to support then `email_support`.

core/email.py
```diff
@@ -209,39 +185,35 @@ def email_to_admin(
         username=None,
         user_email=None,
         cc_user=True,
-        admin_user=None,
```
cdosborn: :+1:

core/email.py
```diff
@@ -209,39 +185,35 @@ def email_to_admin(
...
         request_tracker=False,
         html=False):
     """
     Send a basic email to the admins. Nothing more than subject and message
```
cdosborn: Just a point of commentary. This function's behavior is so needlessly complex. All this if/else bs would be eliminated with a simpler signature:
```
def email_to_admin(subject, body, from, cc=None, html=False)
```

calvinmclean:
Travis is failing because `ATMO_SUPPORT` is not in `local.py` which will be fixed by adding the var to secrets repos

cdosborn:
Travis should generate a valid local.py from its dist variables.

calvinmclean:
ah probably because it is undefined there

calvinmclean:
Failing because of formatting. I'm planning to format with `yapf` after I add tests and finish it up

calvinmclean:
squashed chaotic commits because the direction of the PR changed a few times

api/tests/v2/test_email.py
```diff
@@ -0,0 +1,127 @@
...
+from api.v2.views import (
+    VolumeSupportEmailViewSet, InstanceSupportEmailViewSet,
+    FeedbackEmailViewSet, ResourceEmailViewSet
+)
+from api.tests.factories import (
+    AnonymousUserFactory, InstanceFactory, UserFactory, VolumeFactory
+)
+
+
+class EmailTestHelper(TestCase):
```
cdosborn: Small comment. If this class is just adding methods, you can express this intent in a subtle way by changing the name to `EmailTestMixin` and not having it inherit from `TestCase`. This is meaningful, because your class isn't trying to be something worth instantiating on its own, its just trying to mixin functionality.

api/tests/v2/test_email.py
```diff
@@ -0,0 +1,127 @@
...
+    """
+    This class is used for testing the following ViewSets:
```
cdosborn: Other small note. A class shouldn't mention who uses it, that is bound to become out of date, but more importantly the class doc should just describe what the class itself is responsible for

api/tests/v2/test_email.py
calvinmclean: The reason I had it inherit from `TestCase` was so I could use `self.assertEquals`. I could change the name to `EmailTestMixin` to be more appropriate for its behavior as long as having a mixin inherit from another class is an acceptable practice

api/tests/v2/test_email.py
```diff
@@ -0,0 +1,127 @@
...
+            mock_email_support.assert_called_once()
+            self.assertEquals(response.status_code, 200)
+
+    def anonymous_email_test_helper(self, request, cls, mock_method):
+        view = cls.as_view({'post': 'create'})
+        force_authenticate(request, user=AnonymousUserFactory())
+        with mock.patch(mock_method) as mock_email_support:
+            response = view(request)
+            mock_email_support.assert_not_called()
+            self.assertEquals(response.status_code, 403)
```
cdosborn: Helper methods should not perform tests. When you're reading a test you don't want to trace down assertions in abstractions. 

A test is all about assertions. When you read a test you want to read as little setup code, and only have to make sense about the assertions. It is worth abstracting all the creation and mocking to have simple assertions, but you don't want to hide what assertions are happening in each test.

As a note on style, code duplication is not the worst sin, the worst sin is a poor abstraction. It's a common pattern to avoid code duplication at all costs, this often leads to methods that are so specific or obtuse that they could never be reused, or don't really make sense as a unit of work. By analogy, think about shell aliases and bash functions. You save a bash function when you think it may be useful later, something that is generally useful. You wouldn't make a bash function which searched /var/log/celery/<some node> for the string "cdosborn". If you did you would accrue all these useless functions. Now imagine that you were tasked with understanding someone else's set of functions written this way. It would be hard and not fun (yet ironically this is much of software development). 

Reflecting on that idea, can you appreciate my objection to  `email_test_helper`. I would struggle to come up with a sentence to describe exactly what it does. It's just a clever bundling of statements which prevents code duplication at the cost of being understandable and re-usable. 

Another test for is this abstraction good: do the arguments make sense, could I assume what arguments would need to be passed to this function? Your function `email_admin(subject, body, sender)` is a perfect example of a good abstraction, I wouldn't expect that I would ever need to look inside that function.

cdosborn: Here is an example where I broke out of a ton of the duplicate code into helpful functions at the bottom: `api/tests/v2/test_resource_requests.py`.

Each test reads, blah blah blah `assert` `asert` etc.

calvinmclean:
Tests in `core/tests/test_email.py` are failing and I believe it is because of Celery tasks. I am working on fixing it by mocking `core.email.send_email.si` and it is working so far with `assert_called_once()` and I am now working on testing the `call_args` but didn't have time to finish it today

cdosborn:
Looks great

PR #667, opened 2018-09-12
For application_to_provider, determine correct upload method based on Glance client version specified in main function

c-mart:
## Description

Apparently newer Glance client packages don't even have a v1 attribute, in which case 1. this code throws an `AttributeError` and 2. We're definitely not using old Glance v1. So, don't even test to see if we have a Glance v1 client. This is already hotfixed on J7m prod. 

- [ ] ~Documentation created/updated (include links)~
- [ ] ~If creating/modifying DB models which will contain secrets or sensitive information, PR to [clank](https://github.com/cyverse/clank) updating sanitation queries in `roles/sanitary-sql-access/templates/sanitize-dump.sh.j2`~
- [ ] ~New variables supported in Clank~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,6 +66,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
     ([#655](https://github.com/cyverse/atmosphere/pull/655))
   - Reincluded/fixed broken tests
     ([#657](https://github.com/cyverse/atmosphere/pull/657))
+  - In application_to_provider,py, determine correct upload method based
+    on Glance client version specified in main function
+    ([#667](https://github.com/cyverse/atmosphere/pull/667))
 
 ## [v33-0](https://github.com/cyverse/atmosphere/compare/v32-2...v33-0) - 2018-08-06
 ### Changed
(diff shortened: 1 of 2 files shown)


scripts/application_to_provider.py
```diff
@@ -581,7 +581,7 @@ def migrate_image_data_glance(src_glance_client, dst_glance_client, img_uuid, lo
         with open(local_path, 'rb') as img_file:
             try:
                 # "Upload" method is different for Glance client v1, than for v2
-                if type(dst_glance_client) is glanceclient.v1.client.Client:
+                if hasattr(glanceclient, 'v1') and type(dst_glance_client) is glanceclient.v1.client.Client:
```
cdosborn: Some thoughts.

We call an rtwo method to construct the client for us, how it does that isn't our concern. The package structure of the glance library that rtwo uses has changed, but that shouldn't matter here. This script already gets passed a glance version `dst_glance_client_version`, so it has its own representation of that state, inspecting the class for a version is a redundant representation of that state.

This file actually imports glanceclient, so this script is using an interface that its library no longer supports.

This method above should just be parametrized with the version or the path to the class should be corrected to match its library.

CHANGELOG.md
```diff
@@ -61,6 +61,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
     ([#655](https://github.com/cyverse/atmosphere/pull/655))
   - Reincluded/fixed broken tests
     ([#657](https://github.com/cyverse/atmosphere/pull/657))
+  - In application_to_provider,py, determine correct upload method based
+    on Glance client version specified in main function
+    ([#667](https://github.com/cyverse/atmosphere/pull/667))
```
cdosborn: Doesn't require change, but Its not relevant that the `version is specified in main function` at the change log level. My hope for the CHANGELOG is that its terse and gives you just enough to look into the pr you need.

PR #668, opened 2018-09-12
Fix several issues in attempt to add timeout to tas api

cdosborn:
## Description

The previous pr https://github.com/cyverse/atmosphere/pull/639 did not actually work. 
- `TACC_API_TIMEOUT` was being templated as a string, rather than a number
- The exception raised during a timeout didn't result in offline validation of a users allocation

I tested this locally and fixed the issues. I additionally changed the scope so that the timeout only applies to GETs, I adjusted the variable so that its scope is obvious. Changed `TACC_API_TIMEOUT` to `TACC_READ_API_TIMEOUT`.

  - https://github.com/cyverse/clank/pull/278

### DIFF
--- a/jetstream/tas_api.py
+++ b/jetstream/tas_api.py
@@ -14,10 +14,7 @@ def tacc_api_post(url, post_data, username=None, password=None):
         password = settings.TACC_API_PASS
     logger.debug('url: %s', url)
     # logger.debug("REQ BODY: %s" % post_data)
-    resp = requests.post(
-        url, post_data,
-        auth=(username, password),
-        timeout=settings.TACC_API_TIMEOUT)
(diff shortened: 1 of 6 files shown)


jetstream/tas_api.py
```diff
@@ -30,10 +28,16 @@ def tacc_api_get(url, username=None, password=None):
...
+    try:
+        resp = requests.get(
+            url,
+            auth=(username, password),
+            timeout=settings.TACC_READ_API_TIMEOUT)
+    except ReadTimeout:
+        raise TASAPIException(
+            "TAS API is taking too long to respond, we're timing out ({})".
+            format(url)
+        )
```
cdosborn: It's not clear to me if this is the best place to catch this exception. Open to ideas.

PR #687, opened 2018-09-28
Fix tacc timeout feature

cdosborn:
### Problem
Several tas api endpoints regularly take more than 5 seconds (ex. getting all projects) and were timing out

### Solution
Refactor timeout feature so that the one place that needs a timeout can specify it and handle a timeout exception without affect all other uses of the tas api.

See the commit messages for a longer explanation of the refactor

- [ ] Change to variables in clank


simpsonw:
This is working for me locally and the tests passed after I fixed the merge conflict in the changelog.  Merging PR.

PR #707, opened 2019-06-11
Add hook to create tag from tested commit

calvinmclean:
## Description

Dockerhub is configured to automatically build and test all new pull requests. However, it does not save the images that it creates so we cannot download it for manual testing or atmobeta deployment if we want. This PR introduces a `post_test` hook for Dockerhub. This is just a simple script that will run after the tests to rename and push an image tagged with the branch name and first 7 characters of the commit hash. This will allow us to download and test images from PRs in different states of completion.

Once the Dockerhub builds complete for this PR, you should be able to see a new tag in Dockerhub for each of the commits in this PR (since I modified the changelog after opening the PR):

> atmosphere:dockerhub-hook_58a17c2
> atmosphere:dockerhub-hook_fdb4e9b

**NOTE**: I am open to not including the commit hash if we just want to use the latest version of the PR branch instead of creating many different tagged images.


calvinmclean:
Currently not working despite the identical script working on my fork :thinking:

calvinmclean:
Maybe it does not work because it is running it from my Dockerhub account and restricts access there. This would make sense because someone could hijack our `latest` or version tags by naming their branch something like that. However, this would not be a problem since we append the commit hash. Someone could still use this to create a malicious tag on our Dockerhub repo even if it is not tagged `latest`.

Anyways, this does not work here so I am closing this.

calvinmclean:
Interestingly, the last commit on this was also run as a PR to my fork of Atmosphere and was able to successfully push the image to the **CyVerse** repo. That is really strange.

PR #711, opened 2019-07-01
Update dependencies

simpsonw:
## Description

This pull request contains security updates for Atmosphere's dependencies.  I've chosen to exclude pycrypto from the pull request for two reasons.  First, there isn't a new release of pycrypto that addresses the security vulnerability (apparently there hasn't been a new release of pycrypto for several years).  Second, pycrypto is being used by `jwt.py`, which is a dependency of `cyverse/django-cyverse-auth`, so I feel that it would be more appropriate to either change the dependency in  `cyverse/django-cyverse-auth` or try to fork and do a pull request for `jwt.py`.  It looks like PyCryptodome may be a suitable replacement for pycrypto but again, I don't think it makes sense to add the change here.  (If that's incorrect, just let me know!)


simpsonw:
I updated `jwt.py` to use pycryptodome and was able to test locally.  Everything appeared to work correctly and all the tests passed for me.  I have opened a [a pull request to jwt.py](https://github.com/iPlantCollaborativeOpenSource/jwt.py/pull/2) to replace pycrypto with pycryptodome.

PR #737, opened 2020-05-15
specify setuptool version < 45

zhxu73:
`setuptool` 45.0 and beyond seems to break under python 2, need to specify <45 to work under python2 for now.


zhxu73:
pip does recognize "<45", resubmit as #738

PR #738, opened 2020-05-15
specify setuptools version for python 2

zhxu73:
`setuptool` 45.0 and beyond seems to break under python 2


zhxu73:
reference: https://github.com/pypa/virtualenv/issues/1493

